PR #1, opened 2015-09-11
vw_qgep_cover_update

bogdanvaduva:
Changed vw_qgep_cover_update: modified material -> cover_material, test_views.py: added cover_material in test


m-kuhn:
Thanks a lot

PR #3, opened 2015-10-19
Changed vw_qgep_cover_INSERT and vw_qgep_cover_UPDATE by adding NEW.i…

bogdanvaduva:
Hi m_kuhn you suggested me 2 weeks ago a pull request for setting a default value for identifier.
Sorry for getting back to you so late. Better to be late than ever :D.
Changed vw_qgep_cover_INSERT and vw_qgep_cover_UPDATE by adding NEW.identifier value if null. Also added tests.
I hope it's what you had in mind.
Regards

### DIFF
--- a/test/test_triggers.py
+++ b/test/test_triggers.py
@@ -62,7 +62,7 @@ class TestTriggers(unittest.TestCase, DbTestBase):
         row = self.select('od_structure_part', obj_id)
 
         identifier = row['identifier']
-        assert identifier, "Identifier not set on insert: {}".format(repr(identifier))
+        assert identifier is not None, "Identifier not set on insert: {}".format(repr(identifier))
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_triggers.py
+++ b/test/test_triggers.py
@@ -51,6 +51,7 @@ class TestTriggers(unittest.TestCase, DbTestBase):
         row = self.select('od_structure_part', obj_id)
         assert last_mod != row['last_modification'], "Last modification not set on update of child table (still {})".format(row['last_modification'])
 
+    def test_identifier(self):
         row = {
                 'level': decimal.Decimal('50.000')
         }
--- a/test/test_triggers.py
+++ b/test/test_triggers.py
@@ -62,7 +62,7 @@ class TestTriggers(unittest.TestCase, DbTestBase):
         row = self.select('od_structure_part', obj_id)
 
         identifier = row['identifier']
-        assert identifier is not None, "Identifier not set on insert: {}".format(repr(identifier))
+        assert identifier, "Identifier not set on insert: {}".format(repr(identifier))
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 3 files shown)


test/test_triggers.py
```diff
@@ -51,5 +51,18 @@ def test_last_modified(self):
...
+                'identifier': '',
+                'level': decimal.Decimal('50.000')
+        }
+
+        obj_id = self.insert_check('vw_cover', row)
+
+        row = self.select('od_structure_part', obj_id)
+
+        identifier = row['identifier']
+        assert identifier is not None, "Identifier not set on insert"
```
m-kuhn: Shouldn't it also check that it's not empty? At least there's an empty string specified above which should be replaced by the obj_id and here the check is for None.

Does

    assert identifier, "Identifier not set on insert"

work? In python a string converted to a boolean is False if it's empty or None IIRC.

m-kuhn: Or even better `assert identifier, "Identifier not set on insert: '{}'".format(repr(identifier))`

m-kuhn: Can you do this small update? If the test still passes after that we can merge this important update.

bogdanvaduva:
It doesn't work with assert identifier, .... It has to be assert is not None. At least it is what travis gives me.
Regards

m-kuhn:
Doesn't that mean that the test fails for a valid reason?

On [line 56](https://github.com/QGEP/datamodel/pull/3/files#diff-98e9907bc70fc483e2504f48edf95fd6R56) a new entry is inserted with an empty `''` identifier, on [line 65](https://github.com/QGEP/datamodel/pull/3/files#diff-98e9907bc70fc483e2504f48edf95fd6R65) the check asserts that the new identifier is not `NULL`. So the test succeeds if the identifier is properly updated (=set to the obj_id) but also if nothing at all happens (the identifier is still empty).

My assumption is at the moment is the second case: A new record with an empty identifier is inserted, the trigger does not work for some reason, the check for None reports positive. But it's not what we really want to happen.

bogdanvaduva:
Yes, you were right about your assumption. Here was the problem: I didn't changed vw_cover_insert trigger. Now I did changed vw_cover_insert. I also changed test_trigger.py. 

You have to bear with me, because I am not a python programmer so everything is new.

m-kuhn:
Manually merged in ede1f3ec484d9e65a062ce0b40c617ecd1edf79a

I realized that changes to vw_cover are useless since the file is generated and therefore changes will not be long-lasting. So the only merged change concerns vw_qgep_cover which is manually managed.

If changes to vw_cover should be required as well they'll need to be implemented differently. But if that's required there's probably a lot more to it (i.e. more generic). I hope the current fix is all that's needed.

Thanks a lot @bogdanvaduva!

sjib:
Tell me what I have to change in the Metadatabase export for vw_cover – then I will implement it.

m-kuhn:
I wonder if we should really introduce this kind of magic. For me it's fine at the moment.

The question would be: When inserting a second cover on an existing wastewater structure, should the identifier automatically be synchronized with its obj_id when left empty?
And if this question is answered with yes, then it should probably apply to all objects with identifiers. I think it's an overkill. Therefore if it's for me, leave it as is.

sjib:
I think also this is P2 or P3 – nice to have but not worth a lot of time and money right now.

PR #7, opened 2015-12-02
Added fn_profile, fn_downstream, fn_upstream, fn_reach_for_downstream, fn_reach_for_upstream

bogdanvaduva:
I've added 5 new functions.
fn_profile, 
fn_downstream, 
fn_upstream, 
fn_reach_for_downstream, 
fn_reach_for_upstream.

This way we can discuss the new functions (as @m-kuhn suggested).

Regards,


m-kuhn:
The tests fail because your functions are added before the views.

The order is defined here:
https://github.com/QGEP/datamodel/blob/master/scripts/db_setup.sh

Define your functions in a new .sql script which is executed in the end and everything should be fine.

PR #8, opened 2015-12-03
Added fn_profile, fn_downstream, fn_upstream, fn_reach_for_downstream, fn_reach_for_upstream in a new file

bogdanvaduva:
Added fn_profile, fn_downstream, fn_upstream, fn_reach_for_downstream, fn_reach_for_upstream in a new file called 12_profile_functions.sql

### DIFF
--- a/12_profile_functions.sql
+++ b/12_profile_functions.sql
@@ -326,3 +326,4 @@ ST_Dwithin(ST_StartPoint(qgep.vw_qgep_reach.progression_geometry),
 ) LIMIT 1;$BODY$
   LANGUAGE sql VOLATILE
   COST 100;
+  
--- a/12_profile_functions.sql
+++ b/12_profile_functions.sql
@@ -1,0 +1,328 @@
+-- Function: qgep.fn_profile(text)
+
+-- DROP FUNCTION qgep.fn_profile(text);
+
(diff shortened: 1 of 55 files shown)


3nids:
Why using coalesce in insert triggers?
There would be no chance that ws.obj_id would be null, no?

bogdanvaduva:
Are you refering about qgep.vw_qgep_cover_insert ?

3nids:
sorry, you're right, I read a bit quickly.

m-kuhn:
Can you create a separate pull request for the changes on the views? I think the two issues do not depend on each other.

Also, have you considered using default values instead of coalesce? No strong preference, just a thought since I think it makes the intention more explicit.

    ALTER VIEW vw_wastewater_node ALTER COLUMN obj_id SET DEFAULT qgep.generate_oid('od_wastewater_node');

bogdanvaduva:
Yes, I will create separate pull requests. 
I didn't thought about default values. Those can only be used on materialized views (but it's a more clean approach). Thanks.

bogdanvaduva:
I've modified my datamodel to use default column values for obj_id in views. 
It didn't work because I had inserts with the following format INSERT INTO .. (obj_id, ... ) VALUES (null, ...) which caused error. May be I missed something. Anyway I will make a pull request with COALESCE. 
If you have ideas let me know.
Regards.

PR #9, opened 2015-12-16
Changed db_setup.sh, insert triggers and 03_qgep_db.sql

bogdanvaduva:
I've changed db_setup.sh/03_qgep_db.sql to accept SRID.
I also said before that I will make a pull request for insert triggers (with COALESCE).


m-kuhn:
Can split up these changes?

I think there should be three different pull requests now:

 * changes to the views (insert/coalesce)
 * profile functions
 * changes regarding srid

This way it's possible to review the issues isolated.

You will need to work in different branches if you don't already do.
For the changes to 03_qgep_db.sql, this is an automatically generated file and your changes will be overwritten. It's better to add a "fixup script" that will adjust the SRID at a later stage.

PR #13, opened 2016-03-07
add damage view creation

3nids:
this also updates metaproject submodule which now use a python function instead of plpgsql

also dirty fix for quantification fields (varchar vs integer)


3nids:
@m-kuhn I don't understand, it works well locally. Can you have a quick look please?
Cheers!

sjib:
Hello Denis

Quantification1 and 2 have to be varchar – not all allowed values fit into integer!!!

m-kuhn:
@3nids the travis build is fixed
I think the issue with datatype has to be still sorted out before merging

And nice job with the script invocation in -c

3nids:
ok, thanks a lot!
indeed, the code for view creation is a lot nicer than before!
I will change the quantification type and merge if it's ok for you.
cheers

m-kuhn:
Proceed as you wish

3nids:
thanks @sjib

PR #14, opened 2016-04-15
Update VSA-KEK 20160415

sjib:
Updated VSA-KEK extension

PR #15, opened 2016-04-15
Integrate KEK dictionnaries

m-kuhn:
Based on #14 

Changes:
 * Use utf-8 encoding
 * plantype is qgep.plantype
 * remove previous fixups by @3nids


m-kuhn:
@3nids can you check if the fixups which have been removed are properly replaced by the new SQL script?

3nids:
thanks!

PR #16, opened 2016-04-18
ON DELETE Constraints

m-kuhn:


03_qgep_db_dss.sql
```diff
@@ -3923,7 +3923,7 @@ ALTER TABLE qgep.vl_structure_part_renovation_demand ADD CONSTRAINT pkey_qgep_vl
  REFERENCES qgep.vl_structure_part_renovation_demand (code) MATCH SIMPLE 
  ON UPDATE RESTRICT ON DELETE RESTRICT;
 ALTER TABLE qgep.od_structure_part ADD COLUMN fk_wastewater_structure varchar (16);
-ALTER TABLE qgep.od_structure_part ADD CONSTRAINT rel_structure_part_wastewater_structure FOREIGN KEY (fk_wastewater_structure) REFERENCES qgep.od_wastewater_structure(obj_id);
+ALTER TABLE qgep.od_structure_part ADD CONSTRAINT rel_structure_part_wastewater_structure FOREIGN KEY (fk_wastewater_structure) REFERENCES qgep.od_wastewater_structure(obj_id) ON DELETE set null;
```
m-kuhn: Hi  @sjib 

Is the `ON DELETE SET NULL` action correct for the *structure part* -> *wastewater structure* relation?
I would expect that the structure parts are deleted as well when a wastewater structure is deleted. With `ON DELETE SET NULL` they will be orphaned.

sjib: Hi @m-kuhn 
It is the other way round. When I delete a cover (which is a structure part) I do not want the wastewater structure to be deleted also. E.g. I delete a second cover or just a benching or ladder. So I think it is correct this way. This points from structure_part -> wastewater structure, not the other way round, then it would be wrong

m-kuhn: The `ON DELETE` action does not control what happens to the wastewater structure in case a part is deleted. If the child (part) is deleted, the parent (structure) always stays alive. The `ON DELETE` action controls what happens to the N children when the parent is deleted.

m-kuhn: See also www.postgresql.org/docs/9.5/static/ddl-constraints.html (5.3.5 towards the end of the page)

m-kuhn:
Superseded by #18

PR #19, opened 2016-04-28
[travis] use trusty (faster to start)

3nids:
https://docs.travis-ci.com/user/ci-environment/


m-kuhn:
Is it also faster compared to containerized builds?

3nids:
oh right, forgot about these.

PR #21, opened 2016-08-08
Create and fill in romanian values in the vl_* tables

tudorbarascu:
I have split this into two separate commits:
- the first one generates the ``value_ro`` and ``abbr_ro`` columns into the vl_* tables
 When the ``qgep.is_value_list_base`` will also have these columns then we can do a ``git revert``
- the second one fills in the values

None of them should interfere with Stefan's work and they provide a quicker and easier way for myself (and others) to keep translations up to date.

Thanks!


m-kuhn:
Sounds good to me

PR #22, opened 2016-08-24
New folder interlis

sjib:
Scripts for INTERLIS Export schema and ili2pg INTERLIS export


sjib:
@m-kuhn can you merge this?

m-kuhn:
Thanks!

Should that all go into the datamodel folder?
Or should parts of it go to the plugin?
Should we ship interlis here or can the user just download the latest version himself?
It's quite a big change, do you want some parts to be reviewed?

sjib:
No, I think everything should be together as a package

Updates should be made there with newer version as they might need adaptions

Whether things will go into the plugin I cannot say right now.

m-kuhn:
Ok to bundle it if ili2pg is not API stable between version or depends on a certain model schema.
CC @claeis

----

So far we have only SQL inside the datamodel repository, all logic in the plugin repository and all documentation in the documentation repository. If there is no good reason to change this, I would propose to keep it this way.

claeis:
ili2pg follows normal semantic versioning http://semver.org/

m-kuhn:
Will go to the repo https://github.com/QGEP/QGEP

PR #23, opened 2016-09-12
fix moving covers

oliver-ch:
solves [issue #228](https://github.com/QGEP/QGEP/issues/228)


m-kuhn:
Thanks Oliver.
Looks like for now we need this patch and accept the loss of curves on move.

Looks like ST_SetPoint does not support curves, I know that a group around @andreasneumann is planning to add better curve support to postgis, so we may be able to revert this patch in the future.

PR #29, opened 2017-01-18
Create fix_3d_geometry.sql

varrieta:


m-kuhn:
Integrated in the latest template database. Some views need to be adjusted as well.
Thank you very much!

PR #30, opened 2017-01-18
Update db_setup.sh

varrieta:


m-kuhn:
Can you add this to the other pull request?
I.e. edit the branch `patch-1` here: https://github.com/varrieta/datamodel/blob/patch-1/scripts/db_setup.sh

m-kuhn:
I realized that I'm able to edit your pull request, let's see if it works.

PR #31, opened 2017-01-19
Add inheritance view for organisation

m-kuhn:
@3nids any idea if we can merge the attribute `perimeter_geometry` which is on multiple children (`od_canton`, `od_municipality`). It would be a pity to remap them into mulitple geometry columns.


3nids:
yes, there is the  `merge_columns`option. See the readme for exact usage.

m-kuhn:
> yes, there is the merge_columnsoption. See the readme for exact usage.

nice!

PR #32, opened 2017-02-13
Update fix_3d_geometry.sql

kfischerar:
Moving from 2d to 3d geometry in LV95=2056 (av03=21781)
only tables with real 3d geometry


m-kuhn:
https://github.com/kfischerar/datamodel/tree/patch-1

PR #34, opened 2017-03-31
Set geometry type on vw_qgep_wastewater_structure

m-kuhn:


ponceta:
This will fail if you have 3D geometries! I wouldn't go that way!

Or force2d...

SELECT ST_ASTEXT(ST_FORCE2D(st_collect(co.situation_geometry))::geometry(MultiPoint, 21781)) FROM qgep.od_cover co

But this will be pain for 3D users.

ponceta:
Is there any reason to handle 2D geometries today? Since 3D will be more and more demanded in the future?

m-kuhn:
3D is only activated for some tables (mainly od_reach) `qgep_cover` is not 3D at the moment (the Z part of the coordinate is in the attribute `level`).

The attribute `level` is specified as part of the SIA-405 standard. Adding a Z coordinate will introduce redundancy which should be avoided.

Apart from this I don't have a strong opinion.

ponceta:
My point of view is purely technical and on the db admin point of view.

Working with 2D and 3D geometries simultaneously in QGIS is not well supported.

If you create a Reach (in 3D) snapped from a 2D point, QGIS (in its actual version) will try to create a 2D object which can't be store in the reach table since the database is requiring a 3D.

I discovered this since for QWAT, the technicians were snaping to the 2D polygons of the cadastal layers.

Redondancy on this attribute is a small price to pay compared to the users incomprehension in front of DB error messages :)

But I agree that it demands triggers and a bit of good thinking to update both in a good way.

m-kuhn:
In connection with snapping, the 3D coordinates of a cover are not really useful (they are usually a couple of meters higher than the reach), so it might lead to inappropriate results to snap to the Z value of the cover.

Anyway, if there's an issue with snapping and 3D / 2D, I think we should fix that on QGIS side and not adjust the DB model to work around the issue. I'm happy to help with that :)

PR #37, opened 2017-05-11
add script to generate export views 

3nids:
to be used for data consultation only, such as web clients.

proof of concept with manholes, will go on with pipes.

### DIFF
--- a/export/insert_views.sh
+++ b/export/insert_views.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Exit on error
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+psql -c "DROP SCHEMA IF EXISTS qgep_export CASCADE;"
+psql -c "CREATE SCHEMA qgep_export;"
+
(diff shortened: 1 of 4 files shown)


3nids:
any objection to merge this?

sjib:
Generally not - can you show what the result of this is?

3nids:
reaches, added 

@ṣjib this is a view joining vw_qgep_reach and vw_qgep_wastewater_structure with their vl tables respectively, allowing direct access of values for consultation.

ponceta:
@3nids wouldn't it be great to have a drop views functions to generate the drop views sql instructions on the same basis?

ponceta:
And separate yaml files for better modularity?

m-kuhn:
> @3nids wouldn't it be great to have a drop views function to generate the drop views sql instructions on the same basis?

The drop view statements need to be static.

E.g. a view is created in version 1.1, and removed in 1.3, we still need it in the migration script for 1.4 when migrating from 1.2.

What the generator could do is to add drop views to a growing list of views that could potentially be there.

PR #43, opened 2017-09-08
Update vw_qgep_wastewater_structure.sql

varrieta:
Shouldn't it be ws_remark instead of remark?

### DIFF
--- a/view/vw_qgep_wastewater_structure.sql
+++ b/view/vw_qgep_wastewater_structure.sql
@@ -155,7 +155,7 @@ BEGIN
     , NEW.inspection_interval
     , NEW.location_name
     , NEW.records
-    , NEW.remark
+    , NEW.ws_remark
     , NEW.renovation_necessity
     , NEW.replacement_value
     , NEW.rv_base_year


m-kuhn:
There are quite a few columns that need to be renamed. They should probably be changed all in one go https://github.com/QGEP/QGEP/issues/318

ponceta:
@m-kuhn :+1:

varrieta:
Not sure it is about renaiming the column.

It just that right now we insert in the wastewater structure the field 'remark' that correspond in the view to the main cover remark (main_co_sp.remark).

m-kuhn:
I guess normally you would expect to have the remark on the wastewater structure rather than cover? With the current (cover based) approach, this gets tricky especially with multiple covers.

varrieta:
I guess yes.

It is just that right now we have two columns for the remark in the wastewaterstructure view:
remark that correspond to the main cover remark and ws_remark that correspond to the WS remark. But for now the insert trigger insert the cover remark (so the field remark) in the WS remark instead of the field ws_remark. 

But if the remark is always the cover one, then maybe we should remove the ws_remark from the view otherwise it does get tricky..

Not sure I'm making my self clear.. ^^

m-kuhn:
If I understand correctly: Remarks for covers should be managed only over the cover layer (and not be available on vw_qgep_wastewater_structure)? Sounds perfect to me.

varrieta:
NO, not exactly. What I was saying is that, if we have the column `ws_remark` in the view, then it should be the one introduced in `od_wastewater_structure.remark` by the trigger. 

Because right now we introduce the wrong remark in the wastewater structure table.

ponceta:
I get that too, but according to https://github.com/QGEP/QGEP/issues/318
since ws is the superclass here, 
ws_remark should be renamed remark
and actual
remark should be renamed as main_co_remark

@varrieta provided a nice workaround since actual insert trigger is doing it wrong (modifying main_cover remark **instead** of wastewater structure remark. )

It is correct on the update trigger actually. (see https://github.com/QGEP/datamodel/pull/43/files#diff-2f96799ed8d8629c38873734ead6bde9R397)

If renaming takes place, all should be renamed consequently.

varrieta:
Yes I agree!

ponceta:
Will be fixed with https://github.com/QGEP/datamodel/pull/53

ponceta:
@varrieta could be closed now, https://github.com/QGEP/datamodel/pull/53/files#diff-2f96799ed8d8629c38873734ead6bde9R29 has fixed this.

PR #45, opened 2017-09-12
Improve performance of reach and wastewater structure views

m-kuhn:
See https://github.com/QGEP/QGEP/issues/308


m-kuhn:
The queries were rewritten so they use a subselect instead of windowed functions to avoid having them return multiple rows for the same wastewater structure.
It also improves the handling of wastewater nodes.
Information for a wastewater node will only be shown in the feature form if there is only a single wastewater node for a wastewater structure (most normal manholes). As soon as there is more than one single wastewater node, one needs to use the relation tables

PR #47, opened 2017-10-24
Update slope_per_mill calculation

varrieta:
To avoid a division by zero that makes the database crashes.. (yes, it happend :S )


varrieta:
Done in #55

m-kuhn:
Let's keep this open until merged.
If you write `Fix #47` into the commit message, this will happen automatically.

ponceta:
https://github.com/QGEP/datamodel/pull/57 has fixed it, this can be closed.

PR #51, opened 2017-11-21
Fix ws and ne remark and identifier

ponceta:
Add prefixes to avoid https://github.com/QGEP/QGEP/issues/375


ponceta:
Fix nosetests and squashed commits.

ponceta:
https://github.com/QGEP/datamodel/pull/53 Will fix this

PR #53, opened 2017-12-13
Add prefixes to vw_qgep_reach

urskaufmann:
prefix = table-shortcut, not for main-table (reach and wastewater_networkelement), most important fields at the beginning

### DIFF
--- a/view/vw_qgep_reach.sql
+++ b/view/vw_qgep_reach.sql
@@ -1,25 +1,31 @@
-﻿
+
 DROP VIEW IF EXISTS qgep.vw_qgep_reach;
 
 CREATE OR REPLACE VIEW qgep.vw_qgep_reach AS
 
-WITH active_maintenance_event AS (
-  SELECT me.obj_id, me.identifier, me.active_zone, mews.fk_wastewater_structure FROM qgep.od_maintenance_event me
-  LEFT JOIN
-    qgep.re_maintenance_event_wastewater_structure mews ON mews.fk_maintenance_event = me.obj_id
-    WHERE active_zone IS NOT NULL
-)
+--WITH active_maintenance_event AS (
+--  SELECT me.obj_id, me.identifier, me.active_zone, mews.fk_wastewater_structure FROM qgep.od_maintenance_event me
+--  LEFT JOIN
+--    qgep.re_maintenance_event_wastewater_structure mews ON mews.fk_maintenance_event = me.obj_id
+--    WHERE active_zone IS NOT NULL)
 
-SELECT re.obj_id,
+SELECT 
+    re.obj_id,
     re.clear_height AS clear_height,
+    re.material,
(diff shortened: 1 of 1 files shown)


ponceta:
Nosetest is failing here : https://github.com/QGEP/datamodel/pull/51/files#diff-89f5b165b6f806b9d62a87ef7a0ba02a

ponceta:
Nosetests should be modified here https://github.com/QGEP/datamodel/pull/51/files#diff-89f5b165b6f806b9d62a87ef7a0ba02a

ponceta:
@m-kuhn @haubourg Would it be possible to fix nosetests and merge this?

test/test_views.py
```diff
@@ -55,7 +55,7 @@ def test_vw_qgep_reach(self):
                 'clear_height': 100,
                 'coefficient_of_friction': 10,
                 'identifier': 'pra',
```
ponceta: ws_identifier

test/test_views.py
```diff
@@ -64,7 +64,7 @@ def test_vw_qgep_reach(self):
                 'clear_height': 200,
                 'coefficient_of_friction': 20,
                 'identifier': '10',
```
ponceta: ws_identifier

view/vw_qgep_reach.sql
```diff
@@ -1,25 +1,31 @@
...
-  SELECT me.obj_id, me.identifier, me.active_zone, mews.fk_wastewater_structure FROM qgep.od_maintenance_event me
-  LEFT JOIN
-    qgep.re_maintenance_event_wastewater_structure mews ON mews.fk_maintenance_event = me.obj_id
-    WHERE active_zone IS NOT NULL
-)
+--WITH active_maintenance_event AS (
+--  SELECT me.obj_id, me.identifier, me.active_zone, mews.fk_wastewater_structure FROM qgep.od_maintenance_event me
+--  LEFT JOIN
+--    qgep.re_maintenance_event_wastewater_structure mews ON mews.fk_maintenance_event = me.obj_id
+--    WHERE active_zone IS NOT NULL)
```
3nids: @urskaufmann can you explain why you removed the active_maintenance_event?

urskaufmann: in the vsa-dss-datamodel or in the is_dictionary_od_field, there is no field active_zone. What is the active_maintenance_event? I think in the normal reach-view, there should not be fields of the me-table. If there is a special usecase, there should be used a special view for this case.

m-kuhn: This was introduced for @kfischerar for generating maps for maintenance events.
However, this solution (that is being deleted in here) also led to performance issues, so the proposal by @urskaufmann sounds good to me.

ponceta:
This can be closed since https://github.com/QGEP/datamodel/pull/68 is merged

PR #56, opened 2017-12-20
Add view vw_file

max-trolliet:
Still need the function DELETE. 

The UPDATE and INSERT function run in postgis. However, they seem to contain errors as it is not working in qgis.

### DIFF
--- a/view/vw_file_full.sql
+++ b/view/vw_file_full.sql
@@ -41,7 +41,15 @@ CREATE TRIGGER vw_file_update
   FOR EACH ROW
   EXECUTE PROCEDURE qgep.vw_file_update();
 
+  -- Trigger: vw_file_delete on qgep.vw_file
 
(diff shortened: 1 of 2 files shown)


view/vw_file_full.sql
```diff
@@ -0,0 +1,147 @@
...
+
+-- DROP VIEW qgep.vw_file;
+
+CREATE OR REPLACE VIEW qgep.vw_file AS 
+ SELECT f.obj_id,
+    f.identifier,
+    f.kind AS file_kind,
+    f.object,
+    f.class,
+    dm.kind AS data_media_kind,
```
m-kuhn: @urskaufmann what aliases should be set here (for f.kind and dm.kind)?

max-trolliet:
added delete function

m-kuhn:
did you close this on purpose?

max-trolliet:
No... ^^ i made a mistake here and don't know how to reopen it...

PR #57, opened 2017-12-21
Add calculated_reach_length trigger + update slope calculation

varrieta:
Add a before trigger on qgep.od_reach to calculate the length and save it into length_effective

 Modify slope calculation in vw_qgep_reach to avoid division by zero that makes the database crashes


varrieta:
@m-kuhn This time it worked :)

varrieta:
If it works like that, can anyone commit? :)

06_symbology_functions.sql
```diff
@@ -392,6 +392,9 @@ BEGIN
     WHEN TG_OP = 'DELETE' THEN
       rp_obj_id = OLD.obj_id;
   END CASE;
+  
+  UPDATE qgep.od_reach
```
m-kuhn: Can you add a comment here that explains why this is done (i.e. so we know in 5 years still, that this is done to retrigger the `calculate_reach_length` trigger on update)

m-kuhn:
Good stuff, would be great to add this comment and then ready to merge

varrieta:
Just did, merci Matthias :)

m-kuhn:
Merci Valentine

PR #58, opened 2017-12-21
Add views for od_file od_data_media...

max-trolliet:


max-trolliet:
We need to recover the active object (or selected object) to which the file is to be linked. Any lead on how to get that using only sql?

3nids:
is this superseded by #66 and can be closed?

m-kuhn:
> is this superseded by #66 and can be closed?

Looks like, thanks @max-trolliet  and @3nids

PR #60, opened 2018-01-11
PUM integration

marioba:
Integration of [pum](https://github.com/opengisch/pum) (Postgres Upgrades Manager) into QGEP to manage database upgrades. 
For more information on using [pum](https://github.com/opengisch/pum), please see the project doc.

To do this, the qgep datamodel's python2 code has been refactored to python3. 

I configured Travis (as done in QWAT) to simulate a database upgrade:

- create 3 db:
  - qgep_prod
  - qgep_test
  - qgep_comp
- restore a dump file created from the last datamodel version into qgep_prod simulating the existing production db
- set the baseline version of qgep_prod to 2018.01.05
- run the `db_setup.sh` script into qgep_comp to be used as the reference
- set the baseline of qgep_comp to the current datamodel tag version found in the system/CURRENT_VERSION.txt file
- run pum's `test-and-upgrade` command to upgrade the qgep_prod db, using the delta files with version number greater or equal found in the delta/ directory, after testing the procedure on qgep_test

Pum uses a 3-number version system (x.x.x). In order to make everything work correctly I propose to use versions like _2018.01.11_ for the versions of the qgep datamodels.


m-kuhn:
Nice ! :tada: 

What are the dumps for?

marioba:
@m-kuhn I put in the delta directory a database dump (there are 2 files because one also includes the public schema, the other not) of the datamodel used by the current master of qgep (datamodel @ 74c65a8). I think the best thing, when you start using pum, is to create a new dump with the latest version of the datamodel and start from there with deltas.
At the moment as an example there are two delta files that create and delete a fake table.

m-kuhn:
Ah ok.
Can you move that into a separate directory? It's more testdata (only interesting on travis and for devs) than delta (interesting for all users)

m-kuhn:
Maybe even take the one [with real data inside](https://github.com/QGEP/data/archive/demodata.zip) because it might point out more issues in the deltas?

vpicavet:
Hi @marioba @m-kuhn and @3nids ! Thanks for the work on this issue. Is there still something missing to merge this PR ?

It would be great also to have a clear explanation of the new setup with PUM, and a documentation for the administrators.

3nids:
Following the mail discussion a few days ago:

> La mise à jour vers la version avec les schémas séparés ne se fera pas au travers du PUM pour des raisons pratiques:
> - les changements sont énormes (tous les triggers et vue à réécrire), faire un delta parait peu réaliste
> - cela permettra d'avoir une base déjà par schéma comme version 1.0

The idea is the to integrate PUM when the [table dispatch ](https://github.com/QGEP/QGEP/issues/391) is done. It will be by the end of this week.

@marioba do you have anything ready for the documentation?

marioba:
@3nids the only existing documentation is that on https://github.com/opengisch/pum

PR #63, opened 2018-02-05
add LOGIN to roles

3nids:


haubourg:
Hi Denis, I think those groups should not be connection roles, they are made for receiving actual connection roles. Shouldn't we instead add some fake roles like 
`aQGEPviewer`
`aQGEPadmin` 
granted to corresponding groups?

something like 

````
CREATE ROLE qgep_viewer NOLOGIN NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION;
CREATE ROLE a_qgep_viewer LOGIN NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION;
GRANT  qgep_viewer to a_qgep_viewer;
````
opinion ?

3nids:
yeah, so the file could remain as is.
Basically it's up to the admin to create the user roles and grant them to the proper category.
I don't think we should create login roles in the init script, but maybe along with demo data.

PR #65, opened 2018-04-05
update test/demodata

3nids:
first try with PUM
I will move them to a dedicated sub-repo later on


3nids:
It misses the authentification for qgep-bot to publish the releases

PR #68, opened 2018-04-20
add prefix in views

3nids:
superseds #53


ponceta:
https://travis-ci.org/QGEP/datamodel/builds/369228854#L1911

Oh crap!

This is partly related to https://github.com/qwat/qwat-data-model/issues/199 right?

Creating deltas for views is bad! We already discussed that on qwat and this confirms it.

We urgently need a rewrite view mechanism for qgep! See https://github.com/QGEP/QGEP/issues/365

Otherwise pum serves nothing... (Should be mentionned in pum's readme)

PR #69, opened 2018-05-10
Dump baseline

3nids:
This will solve the issue for data dumps, the version will be in dump.

But it's a bit of an issue that it cannot be present in the structure only (since you cannot dump only specific tables).

What shall we do?

Propose SQL only for model structure and manually append the SQL dump of `qgep_sys.pum_info` ?

That would maybe help bit by lowering the number of files in release.

We would have:
1. structure only SQL (all schema SQL + pum_info)
2. structure + VL data ( 1 + SQL dump of qgep_vl schema)
3. data SQL and dump ( SQL and binary backup of all schemas)

That would give us 4 files instead of 8 currently.


haubourg:
Well, the many dumps formats were at start adressing different use cases:

- a ddl structure + value list only = core database. For those who want to start a qwat with no data
- demo data only 

- structure only + value list dump because it is easier to handle for data restore. A pure plain DDL without compressed big `COPY` blocs can be edited to solve restore issues (that happens for major postgres versions sometimes, I have always been happy to have those parachutes) 

and with two different dumps, custom format, and plain SQL for the sake of a better interoperability between versions.

haubourg:
If I get it right, this does not adds PUM baseline when user creates a new QGEP database using db_setup.sh, it only creates it in dump releases. Is that correct? 
So what is the recommended way to start with QGEP, the script as the documentation mentions or the release dumps? In my opinion, both methods should lead to a baselined database. 
Any thought about that?

3nids:
@haubourg this also done in the init script now

PR #71, opened 2018-06-01
Remove Channels from vw_qgep_wastewater_structure

3nids:
see https://github.com/QGEP/QGEP/issues/405


sjib:
Where is the delta file for this?

m-kuhn:
This doesn't modify the datastructure, views can be dropped and recreated.

PR #72, opened 2018-06-05
Fix special_structure insert function

ponceta:
ss_function instead of ss_special_structure_function

### DIFF
--- a/view/vw_qgep_wastewater_structure.sql
+++ b/view/vw_qgep_wastewater_structure.sql
@@ -222,7 +222,7 @@ BEGIN
              NEW.obj_id
            , NEW.ss_bypass
            , NEW.ss_emergency_spillway
-           , NEW.ss_special_structure_function
+           , NEW.ss_function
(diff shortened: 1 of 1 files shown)


3nids:
thanks a lot, can you add a delta along?

3nids:
mainly being a copy of the SQL file

3nids:
no worries, I'll do it, I have another PR on the same view

ponceta:
Yes, I'm still thinking that deltas on views are an aberration.

I'm trying migration on new datamodel. We'll see if I find some more before doing anything. Just leaving this here as remember note.

m-kuhn:
> Yes, I'm still thinking that deltas on views are an aberration.

I think we are all on the same page here, would be nice to address that at some point

3nids:
integrated in #71

3nids:
https://github.com/QGEP/datamodel/pull/71/commits

ponceta:
@3nids Thanks!

PR #73, opened 2018-06-29
cast geometry types in network tracking views

3nids:


3nids:
@m-kuhn the network_segment.detail has some point in there (around 200 hundreds in the demo data). I failed to understand / fix this for the moment. any hint?

m-kuhn:
Nope, I have also seen this before but never investigated in detail since it worked for its purpose and is only derived data.

PR #74, opened 2018-06-29
 Quarantine implementation for QField Import

signedav:
Insert/update of `qgep_import.vw_manhole` inserts the data into the `qgep_import.quarantine_manhole`.
An insert/update of `qgep_import.quarantine_manhole` tries an insert/update on the wastewater_structure data, cover and reach/reachpoint - on success it deletes the entry from the "quarantine", on fail the data stay there.

If the data in the view of the tables `qgep_od.wastewater_structure`, `qgep_od.cover` and `qgep_od.manhole` handled over the view `qgep_od.vw_qgep_wastewater_structure` are fine they would be updated, no matter if there is a success with the inlets/outlets. `structure_okay` in  `qgep_import.quarantine_manhole` is set to `true`.

An update of the reach/reachpoint is done automatically when there is exactly 1 inlet and 1 outlet in the incoming data and the existing data. If there is a success with inlet,  `inlet_okay` in  `qgep_import.quarantine_manhole` is set to `true`. If there is a success with outlet, `outlet_okay` in  `qgep_import.quarantine_manhole` is set to `true`.

If these three "okays" are set to true, the entry will be deleted from the "Quarantine".

If not, the user has to change the inlet/outlet data manually and change `inlet_okay`/`outlet_okay` to `true`, that the entry in the quarantine is deleted.

Small demo with QGIS, where there is first a wrong structure and no outlet. After adding the outlet the assignment is fine and it's updated in the live table:
geoconceil_demo

Including tests for the mentioned functionalities.


signedav:
Hey @m-kuhn @3nids 

1.
For my tests I need to the current database as the test_data instead of the `test_data/qgep_demodata_1.0.0.dump` 
But when I compare these structures I can see more differences than what I made \*

So if I replace the `test_data/qgep_demodata_1.0.0.dump` with my database basing on `pg_qgep_2 qgep_v1.0.1-5_structure_and_demo_data.backup` I'm afraid the old test won't work anymore.

How should I handle that? Upload a second test-data-dump and change the Travis for that or make a script including my changes and the changes between the two dumps?

\* like for example the aliases in the view `qgep_od.vw_qgep_wastewater_structure`

2.
Additionally there are in both db's still values that I assume it's deprecated: 
To generate the ids there exists a table called `qgep_sys.dictionary_od_table` containing tablename-values with the "od_"-prefix like:
"od_file" while there is only an existing table calles "qgep_od.file". 

Should I change them in the new structure or did I missed something in my understanding?

3nids:
following our discussion:

1. run the test after pum upgrade (so you have demo data in the latest db version)
2. will fix it https://github.com/QGEP/QGEP/issues/431

signedav:
@3nids Please take a look at the following two "hacks" I made and I'm not sure if they are okay for you:
- I checkoutet your 1.0.2 delta files and commited them again, because otherwise PUM wouldn't have used it. I'm pretty sure there would have been a more elegant way. But I guess it makes no difference in the end.
- Because the 1.0.2 delta files needed to be used before my quarantine delta file I created the 1.0.3 version. I hope I did not confuse everything by that. I could change that back again, but I would have to find a way how PUM is using the delta files in the correct order. Maybe with another prefix...

.travis.yml
```diff
@@ -64,5 +64,9 @@ script:
   # Run pum's test and upgrade
   - yes | pum test-and-upgrade -pp qgep_prod -pt qgep_test -pc qgep_comp -t qgep_sys.pum_info -d $TRAVIS_BUILD_DIR/delta/ -f /tmp/qwat_dump -i constraints views --exclude-schema public -v int SRID 2056
 
+  # Run tests on qgep_prod
+  - export PGSERVICE=qgep_prod
+  - nosetests test/test_import.py
```
3nids: I'd be in favor of running all the tests on the demo dump too. Or did you try and did not work? (conflict with existing data?)

13_import.sql
```diff
@@ -0,0 +1,575 @@
+-- create quarantine table
+
+DROP TABLE IF EXISTS qgep_import.manhole_quarantine CASCADE;
+
+CREATE TABLE qgep_import.manhole_quarantine
+(
+  obj_id character varying(16),
+  identifier character varying(20),
+  situation_geometry geometry(Point,2056),
```
3nids: replace `2056` by `:SRID`

13_import.sql
```diff
@@ -0,0 +1,575 @@
...
+  EXECUTE PROCEDURE qgep_import.vw_manhole_insert_into_quarantine();
+
+
+-- create triggerfunctions and triggers for quarantine table 
+
+CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_structure_update() RETURNS trigger AS $BODY$
+DECLARE 
+  multi_situation_geometry geometry(MultiPoint,2056);
+BEGIN
+  multi_situation_geometry = st_collect(NEW.situation_geometry)::geometry(MultiPoint,2056);
```
3nids: idem

scripts/db_setup.sh
```diff
@@ -94,7 +95,7 @@ psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/view/vw_q
 psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -f ${DIR}/view/vw_qgep_reach.sql
 psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -f ${DIR}/view/vw_file.sql
 psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -c "$(${DIR}/view/vw_oo_overflow.py ${PGSERVICE} ${SRID})"
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -c "$(${DIR}/view/vw_oo_organisation.py ${PGSERVICE} ${SRID})"
+psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -c "$(${DIR}/view/vw_oo_organisation.py ${PGSERVICE})"
```
3nids: don't remove the SRID param!

scripts/db_setup.sh
```diff
@@ -105,3 +106,5 @@ fi
 
 VERSION=$(cat ${DIR}/system/CURRENT_VERSION.txt)
 pum baseline -p ${PGSERVICE} -t qgep_sys.pum_info -d ${DIR}/delta/ -b ${VERSION}
+
+psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -f ${DIR}/13_import.sql
```
3nids: add `-v SRID=$SRID`

system/CURRENT_VERSION.txt
```diff
@@ -1 +1 @@
-1.0.2
+1.0.3
```
3nids: no need, 1.0.2 is already the current development version, 1.0.1 was last release

3nids:
@signedav well done! I didn't look at the technical things, I just comment for the PUM / model versioning issues.
We can stick to 1.0.2, you just need to rename you delta fille to `delta_1.0.2_quarantine.post.py`
Just address to few comments and it should be good to go!

13_import.sql
signedav: This does not work:
```
LINE 3:   multi_situation_geometry geometry(MultiPoint,:SRID);
                                   ^
CONTEXT:  invalid type name "geometry(MultiPoint,:SRID)"
```
I don't understand why because the one on line 9 works perfectly through.

.travis.yml
signedav: No conflict. But on the test `test_vw_qgep_reach` of `test_views.py` it makes an update what takes more than 10 Minutes with data. So Travis times out: 
```
$ nosetests
............
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
Locally it takes that long as well. It's still running...

Question is, if we should optimize the rule to increase the performance for that action or if this test is an "unrealistic" use case, we could skip at least in the test with data.

What do you think?

3nids: yes, just skip that test for now

13_import.sql
3nids: that's because it's a function and not some simple code.
you'll have to do something like
```
EXECUTE format($TRIGGER$
CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_structure_update() RETURNS trigger AS $BODY$
+DECLARE 
+  multi_situation_geometry geometry(MultiPoint,2056);
+BEGIN
+  multi_situation_geometry = st_collect(NEW.situation_geometry)::geometry(MultiPoint, %s);
......
......
$TRIGGER$, :SRID);
```

might need a few tries ;)

signedav: Hm... It seems that `EXECUTE format(` is only working in a `DO`-block and in a `DO`-block the variables aren't available as well. I'll continue trying tomorrow and otherwise put everything into python...

3nids: well, nothing prevents to add a `DO` block in the SQL file (that was not possible in older versions)

13_import.sql
```diff
@@ -0,0 +1,575 @@
...
+  IF (NEW.photo1 IS NOT NULL) THEN
+    INSERT INTO qgep_od.file
+    (
+      object,
+      identifier
+    )
+    VALUES
+    ( 
+      NEW.obj_id,
+      NEW.photo1
```
3nids: we might some renaming on the photo filenames?

3nids: ping @m-kuhn

13_import.sql
```diff
@@ -0,0 +1,575 @@
...
+DROP TRIGGER IF EXISTS after_insert_try_inlet_update ON qgep_import.manhole_quarantine;
+
+CREATE TRIGGER after_insert_try_inlet_update
+  AFTER INSERT
+  ON qgep_import.manhole_quarantine
+  FOR EACH ROW
+  EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_inlet_update();
+
+
+CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_outlet_update() RETURNS trigger AS $BODY$
```
3nids: since the two methods are very similar I would be in favor of having the two trigger functions calling a single function with the inlet/outlet parameter.

13_import.sql
```diff
@@ -0,0 +1,575 @@
...
+  FOR EACH ROW
+  EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_structure_update();
+
+
+-- Some information:
+-- 1. new lets 0 - old lets 0 -> do nothing
+-- 2. new lets 0 - old lets 1 -> delete let
+-- 3. new lets 0 - old lets n -> delete lets
+-- 4. new lets 1 - old lets 0 -> create let (future)
+-- 5. new lets 1 - old lets 1 -> update let
```
3nids: is this robust enough?

13_import.sql
```diff
@@ -0,0 +1,575 @@
...
+    WHERE ws.obj_id = NEW.obj_id );
+
+  -- handle inlets
+  IF ( new_lets > 1 AND old_lets > 0 ) OR old_lets > 1 THEN
+    -- request for update because new lets are bigger 1 (and old lets not 0 ) or old lets are bigger 1
+    RAISE NOTICE 'Impossible to assign inlets - manual edit needed.';
+  ELSE
+    IF new_lets = 0 AND old_lets > 0 THEN
+      -- request for delete because no new lets but old lets
+      RAISE NOTICE 'No new inlets but old ones - manual delete needed.';
```
3nids: shall this be translated?

test/test_import.py
```diff
@@ -0,0 +1,324 @@
...
+        self.assertEqual( row['remark'], 'Strassenauslauf')
+
+        # it should be in the live table qgep_od.reach and qgep_od.reach_point
+        cur = self.cursor()
+        cur.execute("SELECT re.material, re.clear_height, rp.level, ws.co_level\
+            FROM {schema}.reach re\
+            LEFT JOIN {schema}.reach_point rp ON rp.obj_id = re.fk_reach_point_from\
+            LEFT JOIN {schema}.wastewater_networkelement wn ON wn.obj_id = rp.fk_wastewater_networkelement\
+            LEFT JOIN {schema}.vw_qgep_wastewater_structure ws ON ws.obj_id = wn.fk_wastewater_structure\
+            WHERE ws.obj_id = '{obj_id}'".format(schema='qgep_od', obj_id=obj_id))
```
3nids: rather than `execute('...{var1}....{var2}...'.format(var1=...,var2=...))` you should prefer using integrated variables handling:
http://initd.org/psycopg/docs/cursor.html#execute
`execute('....%(var1)s....%(var2)I....', var1=..., var2=...)`

3nids: not a big deal for tests, but good to know

test/test_import.py
```diff
@@ -0,0 +1,324 @@
...
+        self.update('vw_manhole', row, obj_id, 'qgep_import')
+
+        # it should be in the live table qgep_od.reach and qgep_od.reach_point
+        cur = self.cursor()
+        cur.execute("SELECT re.material, re.clear_height, rp.level, ws.co_level\
+            FROM {schema}.reach re\
+            LEFT JOIN {schema}.reach_point rp ON rp.obj_id = re.fk_reach_point_from\
+            LEFT JOIN {schema}.wastewater_networkelement wn ON wn.obj_id = rp.fk_wastewater_networkelement\
+            LEFT JOIN {schema}.vw_qgep_wastewater_structure ws ON ws.obj_id = wn.fk_wastewater_structure\
+            WHERE ws.obj_id = '{obj_id}'".format(schema='qgep_od', obj_id=obj_id))
```
3nids: idem

13_import.sql
signedav: Can you precise in what way there is a renaming needed?

13_import.sql
signedav: Thought about that as well. I'm not sure what is nicer to look at. I like it in general as well not to have code "double". But on the other hand there would be a lot of `if-else` when we combine it. I will see what I can do.

13_import.sql
signedav: I don't know, but if we have the possibility this would be nice. Is there anything existing already regarding that?

test/test_import.py
signedav: Haven't knew that. Thanks.

signedav: This does only work for variable names because it creates single quotes around it.
See http://initd.org/psycopg/docs/usage.html#query-parameters
>> Only query values should be bound via this method: it shouldn’t be used to merge table or field names to the query (Psycopg will try quoting the table name as a string value, generating invalid SQL).
So I could do it with `psycopg2.sql` http://initd.org/psycopg/docs/sql.html#module-psycopg2.sql. It's working from psycopg2 ver. 2.7 on. Do you see a problem with that?

13_import.sql
3nids: the only thing I would think of is to create a table `qgep_sys.error_messages` which has the translation, a setting for the lang (probably another table `qgep_sys.settings` (see [qwat](https://github.com/qwat/qwat-data-model/blob/master/system/settings.sql) for instance).
but this is maybe too much for here.
ping @m-kuhn

13_import.sql
3nids: I thought the redundancy was big enough, but you are better judge

test/test_import.py
m-kuhn: `obj_id` is a variable, no?

signedav: Yes. It's only about the `obj_id`?
So: 
```
        cur.execute("SELECT *\
            FROM {schema}.file\
            WHERE object = '%(obj_id)s'".format(schema='qgep_od'), {'obj_id':obj_id})
```
Or with psycopg2.sql like this:
```
        cur.execute(sql.SQL("SELECT *\
            FROM {schema}.file\
            WHERE object = '{obj_id}'").format(schema=sql.Identifier('qgep_od'), obj_id=sql.Identifier(obj_id)))
```
What is the preferable way?

signedav: Sorry as the second I mean the combi of them:
```
        cur.execute(sql.SQL("SELECT *\
            FROM {schema}.file\
            WHERE object = %(obj_id)s").format(schema=sql.Identifier('qgep_od')), {'obj_id':obj_id}))
```

m-kuhn: I think so. The schema is a hardcoded value, which is mostly here for code readability and maintainability. I'd go for option 1, I assume it will have a higher probability for things like prepared statements because it transfers the knowledge of which parts are variable to psycopg2.

m-kuhn: Update sounds good, should also prevent from SQL injection if the schema is ever configurable.

13_import.sql
3nids: I struggled a bit but did not find a solution.
Let's see if something comes out https://stackoverflow.com/questions/51192840/use-a-variable-in-a-postgresql-execute-statement

m-kuhn: Let's go with 2056 for now if there's no easy solution

13_import.sql
signedav: I decided to follow your advice anyway and make one function. The reason is that I belief it makes it easier to beware the consistency, in case we update logical functionalities in future like delete etc. of existing outlets.

test/test_import.py
signedav: Done

13_import.sql
signedav: Still open. Do I need to do something before merging?

13_import.sql
signedav: Still open. What I don't understand is, that when we rename the path to the photo in the db-entry, we have to rename the existing file as well, right? Where should that be done?

13_import.sql
m-kuhn: I'm not sure we want this translation on database level. Might be something to integrate in the QGIS translation. Not a requirement for merging this.

13_import.sql
m-kuhn: you are proposing to do some postprocessing of the filename, not changing the attribute name?

13_import.sql
3nids: Following the reply on SO, here is what works:
```PLpgSQL
SELECT set_config('my.srid', :SRID::text, false);
DO $DO$
BEGIN
EXECUTE format($TRIGGER$
  CREATE OR REPLACE FUNCTION qgep_od.my_trigger_fcn() RETURNS trigger AS
  $BODY$
    DECLARE
      multi_situation_geometry geometry(MultiPoint,%1$s);
    BEGIN
      multi_situation_geometry = st_collect(NEW.situation_geometry)::geometry(MultiPoint,%1$s);
      .........
      .........
      .........
    END;
  $BODY$
  LANGUAGE plpgsql;
$TRIGGER$, current_setting('my.srid'));
END
$DO$;
```

13_import.sql
3nids: I'm just wondering if we should make the file name a bit more unique:
* on QField side: use time stamp
* on QGEP side: use the object ID to rename the file.

But am I the only one to see this as an issue?

13_import.sql
```diff
@@ -359,14 +356,27 @@ CREATE TRIGGER after_insert_try_structure_update
...
 
-CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_inlet_update() RETURNS trigger AS $BODY$
+CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_let_update() RETURNS trigger AS $BODY$
   DECLARE 
+    is_inlet boolean;
+    let_kind text;
     new_lets integer;
     old_lets integer;
 BEGIN
+  is_inlet := TG_ARGV[0];
```
3nids: might be a bit picky, but let's pass directly the string as argument, that would be clearer in the trigger code, i.e. 
`EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_let_update( 'inlet' );` 
instead of 
`EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_let_update( TRUE );`

13_import.sql
```diff
@@ -0,0 +1,536 @@
...
+-- 9. new lets n - old lets n -> manual update needed
+
+CREATE OR REPLACE FUNCTION qgep_import.manhole_quarantine_try_let_update() RETURNS trigger AS $BODY$
+  DECLARE 
+    is_inlet boolean;
+    let_kind text;
+    new_lets integer;
+    old_lets integer;
+BEGIN
+  is_inlet := TG_ARGV[0];
```
3nids: I might be a bit picky, but let's pass directly the string as argument, that would be clearer in the trigger code, i.e. 
`EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_let_update( 'inlet' );` 
instead of 
`EXECUTE PROCEDURE qgep_import.manhole_quarantine_try_let_update( TRUE );`

3nids:
awesome job!
for me it's good to merge, just the photo file name issue, but it can be discussed later

13_import.sql
m-kuhn: I think it's a good proposal, but I don't think it's a requirement here, I think we are already over the estimated time.

13_import.sql
signedav: Done

13_import.sql
signedav: Used that.

signedav:
Thanks @3nids and @m-kuhn for the review!

3nids:
good job!
shall we release now 1.0.2 ?

varrieta:
Thanks Dave for the great job!

A remark based on the informations in 13_import.sql 

> -- Some information:
> -- 1. new lets 0 - old lets 0 -> do nothing
> -- 2. new lets 0 - old lets 1 -> delete let
> -- 3. new lets 0 - old lets n -> delete lets
> -- 4. new lets 1 - old lets 0 -> create let (future)
> -- 5. new lets 1 - old lets 1 -> update let
> -- 6. new lets 1 - old lets n -> manual update needed
> -- 7. new lets n - old lets 0 -> create lets (future)
> -- 8. new lets n - old lets 1 -> manual update needed
> -- 9. new lets n - old lets n -> manual update needed

I wouldn't delete the inlet/outlet for points 2 and 3 because it could be that the field worker just forgot an inlet/oulet. I would rather use a manual update.

signedav:
@varrieta Yes you are right. These comments are wrong.
The behavior is like we defined: 
```
-- 1. new lets 0 - old lets 0 -> do nothing
-- 2. new lets 0 - old lets 1 -> **manual deletion needed (or not, depending on if it's on purpose or not)**
-- 3. new lets 0 - old lets n -> **manual deletion needed (or not, depending on if it's on purpose or not)**
-- 4. new lets 1 - old lets 0 -> **manual creation needed**
-- 5. new lets 1 - old lets 1 -> update let
-- 6. new lets 1 - old lets n -> manual update needed
-- 7. new lets n - old lets 0 -> **manual creation needed**
-- 8. new lets n - old lets 1 -> manual update needed
-- 9. new lets n - old lets n -> manual update needed
```

signedav:
Pushed the corrected comments on master @varrieta

PR #76, opened 2018-07-11
Updates on quarantine implementation

signedav:
**"verified" field in import view**
A boolean "verified" in the view, that is always false on loading QGIS project.
It's for managing in QGIS if this feature is verified yet or not. On reopening QGIS project it should be false again.

**deleted boolean**
in view qgep_import.vw_manhole is default false - when turned to true the entry will be deleted from the wastewater_structure in the live table

**new function vw_manhole_insert_into_quarantine_or_delete()**
 (instead of vw_manhole_insert_into_quarantine()) started by trigger on_mutation_make_insert_or_delete (instead of on_mutation_make_insert)

**functionality with the deleted boolean to delete from live table**

**updated function manhole_quarantine_try_structure_update()**
  . to fail when there is a _depth
  . to calculate co_level when it's null but there is a _depth and a wn_bottom_level
  . to calculate wn_bottom_level when it's null but there is a _depth and a co_level

**tests for calculation and deleted**
(deleted has to be un-commented after merge of PR #77)


3nids:
you need to provide a delta for this since we already released 1.1.0.
Sorry for the burden

signedav:
Yes you are right.

signedav:
Actually after first tests of the customer, we found other small staff that was missing.
I will make one PR with all these things and the deltas for them. I'll trigger you for reviews then again, if okay...

signedav:
I have a problem with the last tests. I'm not sure if they have something to do with these implementations. Probably you @m-kuhn @3nids know more about the source of the problem.

*Situation:*
I update co_level and _depth - wn_bottom_level is empty and should be calculated out of the other two values. This works: the wn_bottom_level is calculated and the co_level is updated. It seems the _depth would then be calculated again with the other two values, but this seems not to work. If I make the exactly same update again, it works...

*Example:*
```
SELECT co_level, wn_bottom_level, _depth FROM qgep_od.vw_qgep_wastewater_structure where obj_id='ch13p7mzMA000071' 
```
Result:
```
90.000;80.000;10.000
```
I update over the view:
```
UPDATE qgep_import.vw_manhole SET co_level=90, wn_bottom_level=NULL, _depth=10 WHERE obj_id = 'ch13p7mzMA000071';
```
Result:
```
90.000;70.000;10.000
```
And when I make the update again, the _depth changes:
```
90.000;70.000;20.000
```

Of course I just could remove the _depth - Test from my tests because they don't have directly to do with the functionality that is tested. But I believe this could be a problem in productive use. If I'm wrong, then tell me. Then I will remove it from the tests...

3nids:
short question before, why not updating directly on qgep_od.vw_qgep_wastewater_structure ?

signedav:
Because it has to go over the quarantine. And there the `bn_bottom_level` or the `co_level` is calculated regarding the `_depth` and the other existing value. 

Anyway there would be the same issue (just without calculation according the depth):
```
SELECT co_level, wn_bottom_level, _depth FROM qgep_od.vw_qgep_wastewater_structure where obj_id='ch13p7mzMA000071' 
```
```
90.000;40.000;50.000
```
```
UPDATE qgep_od.vw_qgep_wastewater_structure SET co_level=90, wn_bottom_level=30 WHERE obj_id = 'ch13p7mzMA000071';
```
Depth should change to 60 but is still the old value:
```
90.000;30.000;50.000
```
So the same update again:
```
UPDATE qgep_od.vw_qgep_wastewater_structure SET co_level=90, wn_bottom_level=30 WHERE obj_id = 'ch13p7mzMA000071';
```
... and we have the correct calculations.
```
90.000;30.000;60.000
```

3nids:
That's because trigger `update_depth()` is not called after update of `qgep_od.wastewater_node` (where `bottom_level` lies) but only after update of `cover` (`level`).
There is 2 solutions:
* add a call to update_depth() directly in the update trigger of `vw_qgep_wasterwater_structure`
* create a new trigger in `wasterwater_node`

I would vote for second approach since it seems cleaner

3nids:
@m-kuhn any comment?

m-kuhn:
Yep, second approach sounds cleaner

signedav:
@3nids with including the changes on master in between (your implementations) my tests succeed now.
Do you approve?

3nids:
yes, I do!

PR #77, opened 2018-07-12
fix reach_point_update trigger

3nids:
avoid recalculanting all reaches all reaches when updating a single reach point

### DIFF
--- a/06_symbology_functions.sql
+++ b/06_symbology_functions.sql
@@ -378,8 +378,10 @@ CREATE OR REPLACE FUNCTION qgep_od.on_reach_point_update()
   RETURNS trigger AS
 $BODY$
 DECLARE
-  rp_obj_id TEXT;
-  _ws_obj_id TEXT;
+  rp_obj_id text;
+  _ws_obj_id text;
+  ne_obj_ids text[];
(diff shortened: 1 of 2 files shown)


06_symbology_functions.sql
```diff
@@ -378,32 +378,41 @@ CREATE OR REPLACE FUNCTION qgep_od.on_reach_point_update()
...
     WHEN TG_OP = 'DELETE' THEN
       rp_obj_id = OLD.obj_id;
+      ne_obj_ids := ARRAY[OLD.fk_wastewater_networkelement];
   END CASE;
 
 
   UPDATE qgep_od.reach
-  SET progression_geometry = progression_geometry; --To retrigger the calculate_length trigger on reach update
+    SET progression_geometry = progression_geometry
+    WHERE fk_reach_point_from = rp_obj_id OR fk_reach_point_to = rp_obj_id; --To retrigger the calculate_length trigger on reach update
```
3nids: @m-kuhn does it make sense to you?

m-kuhn: yes

ponceta:
Thanks for the fix!

PR #79, opened 2018-08-07
Add default value for vw_wastewater_node

m-kuhn:
@3nids I guess this also requires a delta?


3nids:
yes!

3nids:
@m-kuhn shall I merge?

PR #86, opened 2018-12-13
Also show wastewater nodes in wastewater_structure view

m-kuhn:


m-kuhn:
@urskaufmann mostly interested in your feedback to this modification. It shows the wastewater nodes in addition to covers on the wastewater_structure view. This should help for cases where there is no cover on a wastewater structure.

urskaufmann:
To install I had do delete the "Grant all"-lines with roles that I do not have (I have just role qgep and postgres) and I had to change 217xx to 2056.
Then it worked: there are also point for structure, where there are no covers.

view/vw_qgep_wastewater_structure.sql
```diff
@@ -105,7 +105,7 @@ CREATE OR REPLACE VIEW qgep_od.vw_qgep_wastewater_structure AS
 
   FROM (
     SELECT ws.obj_id,
-      ST_Collect(co.situation_geometry)::geometry(MultiPoint, :SRID) AS situation_geometry,
+      st_collect(st_accum(wn_1.situation_geometry) || st_accum(co.situation_geometry))::geometry(MultiPoint, :SRID) AS situation_geometry,
```
ponceta: `wn` instead of `wn_1`

ponceta:
Fails test_import.py, https://travis-ci.org/QGEP/datamodel/builds/468909885#L2268

@urskaufmann. @m-kuhn I think it is possible to find a solution to label only nodes or covers.

QGIS has the option to avoid label superposition too, right?

3nids:
ping @m-kuhn

ponceta:
We have to show only the "reference" wastewater_node in vw_qgep_wastewater_structure as a single point entity.

>  We have to extend vw_cover and vw_node with several new fields for being able to label and to configure the styles.

@urskaufmann I do agree, what are the actual missing attributes in the view_cover and vw_wastewater_node?

I think only the main cover altitude should be displayed on the wastewater structure label right?

The same way as proposed [here](https://github.com/QGEP/datamodel/pull/112) for nodes

Something tricky will be to display covers and wastewater nodes labels that don't appear already in wastewater structure nodes.

ponceta:
This can be dropped since https://github.com/QGEP/datamodel/pull/111 is now active.

PR #93, opened 2019-03-01
Default value for vw_cover view

m-kuhn:
Fix #477

Band aid for this but.
As soon as we get the green light on customization, this will be applied to all views.


ponceta:
https://github.com/QGEP/QGEP/issues/477

Thanks for the fix!

PR #95, opened 2019-04-09
use pirogue to create views

3nids:
This PR enables (semi-)automatic view creation to allow supporting column addition in base tables.

A new Python package called Pirogue has been created (and release on [Pypi](https://pypi.org/project/pirogue/).
Pum has also been upgraded to support variables in Python delta.

Some remarks:
* Old metaproject submodule has been removed.
* Pirogue is used either to create the join and merge views using dedicated method.
* Travis is green, upgrade path works fine.

To be checked later:
* Should be working but not tested: model updates after 1.3.0 (can be tested as soon as this gets merged and released)
* project has been open, only 3 tables need to be renamed (see https://github.com/QGEP/QGEP/pull/491). But no control has been performed on data editing in QGIS (also waiting on merge to get data dumps). But editing tests are passing on Travis.

***

Pull request pending for documentation: https://github.com/QGEP/docs/pull/72
(traduction can be done after merge)

***

Columns renamed:
* vw_qgep_maintenance.maintenance_type => maintenance_event_type

Views renamed:
* vw_overflow_prank_weir, leapingweir, pump => vw_prank_weir,...


urskaufmann:
1.  Structure part cover has a special position. For other structure parts there are many fields, there will be many fields, that belongs only to one subclass. I am not sure, if this works.  I think its easier the way we have it no.
2. Usage_current and function_hierachic are fields in the channel subclass and are already in vw_channel. vw_channel has all fields of class wastewater_structure and subclass channel. _label and _depth are fields that make only sense with "point"-wastewater-structures (manhole, specialstructure etc.)

3nids:
second question, `wastewater_structure.fk_dataowner` and `wastewater_structure.fk_provider` are missed in reach view. Is this expected?

urskaufmann:
fk_dataowner and fk_dataprovider: in vw_qgep_reach are now the fields of the table wastewater networkelement. The question is, if it is necessary to define different dataowners/providers for wastewater structure and wastewater networkelement.  My opinion: for "normal" datasets, it is not necessary to have so many dataowner/provider fields in the model. I prefer to have only the wastewater structure-dataowner/provider in the vw_qgep_* views. 
So the fields ne.fk_dataowner/provider should be changed to ws.fk_dataowner/provider as ws_fk_dataowner/provider

view/create_views.py
```diff
@@ -0,0 +1,83 @@
...
+    Merge(safe_load(open("view/vw_oo_overflow.yaml")), create_joins=True, variables=variables, pg_service=pg_service).create()
+    Merge(safe_load(open("view/vw_oo_organisation.yaml")), pg_service=pg_service).create()
+
+    run_sql('view/vw_catchment_area_connections.sql', pg_service, variables)
+    run_sql('view/vw_change_points.sql', pg_service, variables)
+    run_sql('view/vw_qgep_import.sql', pg_service, variables)
+
+
+if __name__ == "__main__":
+    pg_service = os.getenv('PGSERVICE')
```
lbartoletti: Why not using args?

view/vw_oo_overflow.yaml
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
lbartoletti: python3

lbartoletti: Hmm... useless

3nids: fixed

view/create_views.py
3nids: fixed

lbartoletti:
@3nids Seems ok for the logic, but please test with the qgis project, I noticed some issues https://github.com/QGEP/QGEP/pull/491 (and fix travis :) )
Also what happens if we try to migrate from 1.2.0/1 since the delta is deleted?

3nids:
failing test is apparently due to recent changes in PUM from @elemoine. Will try to see.

3nids:
failing test was due to recent changes in PUM, fixed now.
Following comment in https://github.com/QGEP/QGEP/pull/491 I added a refresh of materialized views in post-all.py
Regarding the delta, since it's a view handled in pirogue now, there is no reason to keep the delta. So no issue there.
Testing has been done in the project.
Can we expect more in-depth testing on your side?

lbartoletti:
@3nids OK for me. Thanks

PR #88, opened 2019-02-01
Import only verified element

varrieta:
To send only elements that where verified on the field to quarantaine.

PR #100, opened 2019-05-23
The Return of "Import only verified element"

signedav:
Same implementations like in #88

\+ integrated `verified` to tests...


signedav:
Thanks @varrieta and @3nids

PR #101, opened 2019-05-27
remove channel when no reach are left

3nids:
this removes a channel when no reach is left
done using a new trigger


delta/delta_1.3.1_delete_channel_when_no_reach.sql
```diff
@@ -0,0 +1,27 @@
...
+DECLARE
+  channel_id text;
+  reach_count integer;
+BEGIN
+  -- delete channel if no reach left
+  SELECT COUNT(fk_wastewater_structure) INTO reach_count
+    FROM qgep_od.wastewater_networkelement
+    WHERE fk_wastewater_structure = OLD.fk_wastewater_structure;
+  IF reach_count = 0 THEN
+    RAISE NOTICE 'Removing channel (%%) since no reach are left', OLD.fk_wastewater_structure;
```
m-kuhn:
```suggestion
    RAISE NOTICE 'Removing channel (%%) since no reach is left', OLD.fk_wastewater_structure;
```

06_symbology_functions.sql
```diff
@@ -376,6 +376,31 @@ BEGIN
...
+DECLARE
+  channel_id text;
+  reach_count integer;
+BEGIN
+  -- delete channel if no reach left
+  SELECT COUNT(fk_wastewater_structure) INTO reach_count
+    FROM qgep_od.wastewater_networkelement
+    WHERE fk_wastewater_structure = OLD.fk_wastewater_structure;
+  IF reach_count = 0 THEN
+    RAISE NOTICE 'Removing channel (%) since no reach are left', OLD.fk_wastewater_structure;
```
m-kuhn:
```suggestion
    RAISE NOTICE 'Removing channel (%) since no reach is left', OLD.fk_wastewater_structure;
```

06_symbology_functions.sql
```diff
@@ -376,6 +376,31 @@ BEGIN
...
+  RETURNS trigger AS
+$BODY$
+DECLARE
+  channel_id text;
+  reach_count integer;
+BEGIN
+  -- delete channel if no reach left
+  SELECT COUNT(fk_wastewater_structure) INTO reach_count
+    FROM qgep_od.wastewater_networkelement
+    WHERE fk_wastewater_structure = OLD.fk_wastewater_structure;
```
m-kuhn: This will also delete wastewater structures (manholes etc.) when the last wastewater node is removed, no?

sjib: Add a warning if this is happening!

3nids: I don't think so, let's try to write it down:

1. Trigger is on delete on a network element 
2. It counts the number of network elements having a reference to the same wastewater structure (i.e. channel)
2. Deletes channel and wastewater structure having obj_id as the channel one.

On 1., shall I check that it's indeed a reach (and not a reach point)?

3nids:
here is a new version that happens on reach delete.
I had to transform the existing rule to a trigger to make this working.
The renaming of the triggers makes them be executed in proper order (PostgreSQL runs triggers alphabetically and not by order of creation)

06_symbology_functions.sql
m-kuhn: The proposed check sounds good.
So just for the record: a networkelement can be a reach or a wastewater node.

PR #104, opened 2019-07-25
add not null constraints to geometry on basic tables

3nids:
reach_point, wastewater_node, cover and reaches
fixes https://github.com/QGEP/QGEP/issues/510

@sjib @urskaufmann works for you?


3nids:
Tests are finally OK.

I had to drop a few lines from the original dump since it contained some lines without geometry. Is this OK for you @urskaufmann ?
I chose to apply a delta rather than creating a new dump, this was easier.

@sjib do you want to have a look?

sjib:
Do you mean data without geometry in the demo project?

urskaufmann:
I have no problems with the demodata.
What happens, when I move a reach-vertexpoint and the result is a NULL-geometry? Warning? Refuse of the moving? Delete of reach and channel?

In my opinion, the reach-point have always a valid geometry if the reach has a valid geometry. So reach, cover and node are the important tables. Not sure about the catchment area.

3nids:
> when I move a reach-vertexpoint and the result is a NULL-geometry
how do you expect this to happen?

urskaufmann:
> > when I move a reach-vertexpoint and the result is a NULL-geometry
> 
> how do you expect this to happen?

with the vertex tool. If you are not careful with snapping, its possible to snap to the same node as the other end of the reach is already snapped to

3nids:
@urskaufmann sorry I fail a bit to picture the thing. But anyway, is it a correct situation? If not, this PR would help by enforcing the geometry to be not null.

urskaufmann:
@3nids  it is not a correct situation and it is good to enforce the geometry to be not null. My question was: what happens when a NULL-geometry should be saved.

3nids:
I guess you will get an error and will not be able to save. @m-kuhn do you confirm this behavior with transaction mode?

m-kuhn:
> what happens when a NULL-geometry should be saved.

If saving a NULL-geometry is a valid use case these constraints should not be added (surprise, surprise)
If the question was "what happens if the system tries to do so", I assume it will raise an error and not do it. But it needs proper testing.

One workflow that needs particular attention is adding a cover from a wastewater structure form. The form does not yet allow to digitize the geometry directly (pending in https://github.com/qgis/QGIS/pull/30905 ). Does this still work properly?

m-kuhn:
We suggest moving this to more real world testing.

@varrieta and @urskaufmann , would you mind doing this, by applying https://github.com/QGEP/datamodel/blob/bcbd26b5554d342d39c410ed0c2bb3c925387ee9/delta/delta_1.3.2_geometry_not_null_constraints.sql on your databases?

urskaufmann:
Did some short tests (also with split and merge of reaches).
Got once one error (see below), but can not reproduce and don't know the exact steps I made before. For me this change seems ok. 



Layer vw_qgep_Haltung: PostGIS-Fehler beim Attributändern: FEHLER: Einfügen oder Aktualisieren in Tabelle »reach_point« verletzt Fremdschlüssel-Constraint »rel_reach_point_wastewater_networkelement« DETAIL: Schlüssel (fk_wastewater_networkelement)=() ist nicht in Tabelle »wastewater_networkelement« vorhanden. CONTEXT: SQL-Anweisung »UPDATE qgep_od.reach_point SET elevation_accuracy = NEW.rp_from_elevation_accuracy , identifier = NEW.rp_from_identifier , level = NEW.rp_from_level , outlet_shape = NEW.rp_from_outlet_shape , position_of_connection = NEW.rp_from_position_of_connection , remark = NEW.rp_from_remark , situation_geometry = ST_StartPoint(NEW.progression_geometry) , last_modification = NEW.rp_from_last_modification , fk_dataowner = NEW.rp_from_fk_dataowner , fk_provider = NEW.rp_from_fk_provider , fk_wastewater_networkelement = NEW.rp_from_fk_wastewater_networkelement WHERE obj_id = OLD.rp_from_obj_id« PL/pgSQL-Funktion qgep_od.ft_vw_qgep_reach_update() Zeile 24 bei SQL-Anweisung

urskaufmann:
I realized, that **not null** is true also for a line with two points with the same coordinates (length = 0).

m-kuhn:
@urskaufmann `NOT NULL` tests for "is there any data in there" so called "empty geometries" are still possible.

urskaufmann:
I realized this. Empty geometrie is for me as dummy user the same as not null: I don't see something on the screen. So I wonder why this should be allowed. Or why I don't get a warning if I produce a empty geometrie.

m-kuhn:
I agree that from a user perspective there is no difference.
The possibilities we have:

1. NOT NULL CONSTRAINT
2. CHECK CONSTRAINT on ST_IsEmpty
3. CHECK CONSTRAINT on ST_IsValid (checks for invalid geometries like self intersections)
4. Other advanced check constraints about duplicate vertices, bounding boxes etc

Which level would you propose to target?

urskaufmann:
First tought: 1+2
But: if I export the data to INTERLIS, duplicate vertices in reaches are errors. So why not check this when digitizing? There are options in QGIS - layer properties - digitizing to eliminate double vertex or to check geometry is valid. Is it enough to use these option? I didn't test it so far.

urskaufmann:
Made some tests with vw_qgep_reach. I think the options Remove duplicate nodes and Is Valid in Layer Properties - Digitizing should be on. Then 1+2 are ok.

m-kuhn:
@3nids as author and @ponceta as initiator of this pull request, what's your opinion?

Should we first proceed with this one and discuss a followup "is empty" check later on or directly test the "is empty"?

ponceta:
I tested the not NULL approach. I would be in favor of merging this and continue on "is empty" in another pull request.

PR #105, opened 2019-07-30
Add rights and default permissions to qgep_export schema

ponceta:
Correct rights to qgep_export.vw_export_reach and qgep_export.vw_export_wastewater_structure will be set correctly at each generation


3nids:
why do we need to give some rights to postgres, that should be left to the admin no?

3nids:
thanks

PR #109, opened 2019-09-06
increase qgep_od.maintenance_event.base_data length

3nids:
for better flexibility to save report file path


tproduit:
When I try to update my database with PUM I get the following error:
```
cannot alter type of a column used by a view or a rule
DETAIL: rule _RETURN on view qgep_od.vw_qgep_maintenance depends on column "base_data"
```

3nids:
Normally the views should be dropped when you use PUM.
Now I recall the issue:
* views should be dropped before running the delta
* we cannot do it in QGEP because old detlas (<1.2) are actually playing with views/triggers. So if you drop the views, the deltas will fail
* hence the drop_views.sql should be run only if we are above 1.2 as current version

The only path I see would be the following:
1) for 1.3.x make it work, by manually specifying the view to delete in the delta
2) release 2.0, from which we introduce the drop_views.sql call as a pre-all delta

@m-kuhn works for you?
@tproduit could you try by copying the file view/drop_views.sql to the delta folder and rename it pre-all.sql.

tproduit:
@3nids it seems to work. However an error is raised, pum detects a difference in the trigger on_wastewaternode_change() but looks like a false alarm.

PR #113, opened 2019-10-16
Fix typo in GRANTS

ponceta:
qgep_manager does not need to be defined as GRANTED for qgep_od since he inherits of qgep_user which is already allowed to access.

But he needs to be GRANTED for qgep_vl TABLES along with qgep_vl SEQUENCES.


ponceta:
Fix https://github.com/QGEP/QGEP/issues/525

ponceta:
Thanks!

PR #115, opened 2019-11-08
Remove vw_qgep_reach.fk_wastewater_structure

ponceta:
This is an exact copy of vw_qgep_reach.ws_obj_id, so there is no reason to keep it.


ponceta:
This fix the cutting of a correctly connected reach with QGIS tool using QGIS, avoiding data loss when cancelling.

@m-kuhn this can be merged safely

m-kuhn:
One important thing to note on this pull request:

The attribute `fk_wastewater_structure` from the view !`w_qgep_reach` has been removed. It was redundant with `vw_qgep_reach.ws_obj_id` which should be used as a drop in replacement wherever `vw_qgep_reach.fk_wastewater_structure` was used.

PR #116, opened 2019-11-21
Fix OID for covers and nodes

ponceta:
Fix https://github.com/QGEP/QGEP/issues/519

@3nids why is the alter view defined 2 times?

This avoid having different obj_id when editing from 

qgep_od.vw_qgep_wastewater_structure
qgep_od.vw_qgep_wastewater_node
qgep_od.vw_qgep_cover

### DIFF
--- a/view/vw_qgep_wastewater_structure.py
+++ b/view/vw_qgep_wastewater_structure.py
@@ -94,8 +94,8 @@ def vw_qgep_wastewater_structure(srid: int,
         {extra_joins};
        
         ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_structure');
-        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER co_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','structure_part');
-        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER wn_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_node');
+        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER co_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','cover');
+        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER wn_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_networkelement');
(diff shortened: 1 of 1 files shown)


view/vw_qgep_wastewater_structure.py
```diff
@@ -94,8 +94,8 @@ def vw_qgep_wastewater_structure(srid: int,
         {extra_joins};
        
         ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_structure');
-        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER co_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','structure_part');
-        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER wn_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_node');
```
m-kuhn: I think the second one needs to stay as "wastewater_node". As far as I understand, we want the obj id of the more specific type.

* cover is more specific than structure_part
* wastewater_node is more specific than wasterwater_networkelement

ponceta: Ok, then I need support on this.

I tried defining WN but it ended with WE prefix for vw_qgep_ws generated WN

Maybe linked to https://github.com/QGEP/QGEP/issues/536

Can you or @3nids have a look on that on Pully's behalf. It would be great to fix that quickly.

3nids: indeed, second line is not need anymore since e4f5c8c6982467beb24489a007abdb6a3007a8d2
I will drop it and merge

view/vw_qgep_wastewater_structure.py
```diff
@@ -94,8 +94,8 @@ def vw_qgep_wastewater_structure(srid: int,
...
+        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER co_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','cover');
+        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER wn_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_networkelement');
```
3nids:
```suggestion
        ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER wn_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_node');
```

ponceta:
Already solved by
https://github.com/QGEP/datamodel/commit/016ac9c6560c32258fe9c698346cce4c7019c4df

PR #118, opened 2019-12-02
Single point for wastewater structures

m-kuhn:
This takes the wastewater node (where present) or the cover (as fallback) for the wastewater geometry on vw_qgep_wastewater_structure.

**Warning:** this modifies the geometry type of vw_qgep_wastewater_structure from MultiPointZ to PointZ, the qgis project will need to be update accordingly. If you have a local project to patch, open it in an editor and search replace MultiPointZ with PointZ.

**Warning 2:** this will take the Z value provided in the insert for "wastewater node" instead of "cover". The "cover" level must now be filled with the appropriate `co_level` attribute.

Fix qgep/QGEP#532

CC @ponceta


m-kuhn:
One open question, the Z value of the new geometry is taken from either wastewater node or cover (if the wastewater node is missing). Should the Z value be set to NULL in the view if the wastewater node is missing?

sjib:
My opinion: I would not mix information from two objects (cover and wastewater node). So if you take the geometry of cover then you should also take its z value. You should only set it to NULL if there is no cover.level

m-kuhn:
What exactly are the objects without wastewater node and only covers?
@urskaufmann you had an example

m-kuhn:
This still needs to be checked carefully for proper behavior.

**Previously** the altitude of a new wastewater structure was given to the **cover**.
If I understood @ponceta correctly, the idea is to use these points as snapping targets for reaches, meaning the "underground part", so the Z values should come from **bottom level**.

I have left a proposal below, please comment or like/dislike as you feel is appropriate

# Proposed behavior

## Select wastewater structure:
- If wastewater node is present, this one will be used (including Z) -> important for snapping
- If wastewater node is absent, cover will be used (without Z) -> better not set the Z value for a reach from a snapped cover level

## Create a new wastewater structure:

### Wastewater node:
- if wn_bottom_level = NULL -> point Z will be taken (wn_bottom_level will be set to Z)
- if wn_bottom_level IS NOT NULL -> point Z will be taken from wn_bottom_level

### Cover:
- will be taken from co_level attribute. Only.

## Update a wastewater structure:

- Z changes will only be propagated from the attributes only
- (or should Z changes on the geometry be propagated to the bottom level too?)

urskaufmann:
Example: Manhole on a pump-pressure-pipe for maintenace use (z.B. ein Spülstutzen oder ein Be-Entlüftungsventil): there is no hydraulic connection between the Manhole and the pipe, so there is no need for a networknode.

urskaufmann:
about the Proposal (thanks Matthias)
Select: Reaches snap to the wastewater_node theme, not to the wastewater-structure theme. The z-value for reach/reachpoints is never taken from a cover level.
Create / node:
"if wn_bottom_level = NULL -> point Z will be taken (wn_bottom_level will be set to Z)": I don't see exactly what you mean. What is clear: if you have measured points (e.g. GPS), you must be very careful what the z-value means: cover level, pipe crown(top of the pipe) or bottom level/bottom of the pipe or terrain level. I will work with attributes, not with 3d-points.
Update: only attributes is ok

m-kuhn:
Thanks Urs

> "if wn_bottom_level = NULL -> point Z will be taken (wn_bottom_level will be set to Z)"

If the wn_bottom_level attribute is specified in the dialog, this attribute will be written to node.bottom_level and Z, if it is left empty, the Z value from the wastewater structure point will be written to node.bottom_level and geometry.

ponceta:
> If the wn_bottom_level attribute is specified in the dialog, this attribute will be written to node.bottom_level and Z, if it is left empty, the Z value from the wastewater structure point will be written to node.bottom_level and geometry.

Yes, that's the User's priority.

User has a Z suggested from the QGIS default value for this field (issued from snapping on node layers), which he can change to update both bottom_level and Z value of geometry.

m-kuhn:
@ponceta in a discussion with @urskaufmann we also considered an option where wastewater structure would have no Z at all (i.e. wastewater structure points would be 2D).

Reasons:

- There is no clear definition of what the Z value in wastewater structures would be
- Snapping must be done on wastewater nodes anyway, as there can be multiple wastewater nodes in a single wastewater structure and if you want to snap to a secondary node, they won't be on the wastewater structure layer in the future
- Different altitude values are available in attributes, a context in which their meaning is precisely defined

Would you be ok with that as well?

ponceta:
As we discussed with @m-kuhn and as a short summary :

 - I'm not strongly against 2D geometries for wastewater if they are created from mixed (wastewater_node / cover if no wastewater node) *

 - As I told @m-kuhn QGIS should be fixed to accept 3D point geoms in 2D point Layers (converting 3D geometries to 2D) (vw_qgep_wastewater_structure is the default insert view)

- 3D geometries of wastewater structures is not fully usable yet since a change in Z will not be set as a bottom_level update (the user has to set the 3D geometry of the underlying wastewater node)

 - Snapping for wastewater structures is mainly done on construction points (3D layer)
 - Snapping for reaches is indeed getting nodes from wastewater_nodes or interpolated values of reaches vertices (when there is no underlying node)

\* I'm not a fan of wastewater without node with our current cadastral definition of the node being the bottom of the wastewater structure.

m-kuhn:
Thank you for the summary @ponceta 

The current status is that it works for the core model (not yet heavily tested), but there are implications on the import views as well which makes this more complicated than originally expected and we will need to continue investigating these issues to see how to continue.

I will have a look at this with @signedav first half next week, so we can make a decision if we will wait for this for the next model release or if we need to postpone this and make a release without it first to ship the SWMM datamodel and other important updates.

m-kuhn:
The current status is:

 - The view `qgep_od.vw_qgep_wastewater_structure` is now 2D only, defined by the coordinates of either the main wastweater node - or if not defined the main cover
 - Altitudes will only be assigned from the attributes `wn_bottom_level` / `co_level`, explicitly

Is everyone fine with this approach?
Is someone able to test this in practice?

@3nids I would be happy to get a review from you and to do a release soon (this year still).

PR #120, opened 2019-12-02
Improve error messaging when SRID is missing

m-kuhn:
Currently if `-v int SRID 2056` is missing, it will fail with the following cryptic message

> Applying post-all.py... int() argument must be a string, a bytes-like object or a number, not 'NoneType'

### DIFF
--- a/delta/post-all.py
+++ b/delta/post-all.py
@@ -14,6 +14,8 @@ class CreateViews(DeltaPy):
         qgep_wastewater_structure_extra = self.variables.get('qgep_wastewater_structure_extra', None)
         qgep_reach_extra = self.variables.get('qgep_reach_extra', None)
 
+        if not self.variables.get('SRID'):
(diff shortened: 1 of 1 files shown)


delta/post-all.py
```diff
@@ -14,6 +14,8 @@ def run(self):
         qgep_wastewater_structure_extra = self.variables.get('qgep_wastewater_structure_extra', None)
         qgep_reach_extra = self.variables.get('qgep_reach_extra', None)
 
+        if not self.variables.get('SRID'):
+            raise RuntimeError('SRID not specified. Add `-v int SRID 2056` to the updgrade command.')
```
3nids:
```suggestion
            raise RuntimeError('SRID not specified. Add `-v int SRID 2056` (or the corresponding EPSG code) to the upgrade command.')
```

PR #125, opened 2020-02-04
fix delete file trigger function

3nids:


ponceta:
It solves the problem! Thanks!

PR #126, opened 2020-02-21
Enforce obj_id generation for vw_file

ponceta:
Requires "Provider default values evaluation" activated on QGIS to run smoothly.


3nids:
thanks @ponceta

PR #127, opened 2020-02-25
improve labelling

olivierdalang:
As per discussion on https://github.com/QGEP/QGEP/issues/524#issuecomment-543209174


05_data_model_extensions.sql
```diff
@@ -20,6 +20,19 @@ added solely for QGEP';
 ALTER TABLE qgep_od.wastewater_structure ADD COLUMN _label text;
 COMMENT ON COLUMN qgep_od.wastewater_structure._label IS 'not part of the VSA-DSS data model
 added solely for QGEP';
+ALTER TABLE qgep_od.wastewater_structure ADD COLUMN _cover_label text;
+COMMENT ON COLUMN qgep_od.wastewater_structure._cover_label IS 'not part of the VSA-DSS data model
```
m-kuhn: Can you add some description of the content to this comment (what it's generated from and what is in there), the VSA-DSS parts are in my opinion mostly a comment for the comment ;)

m-kuhn:
Looks good to me

olivierdalang:
I added some tests and think this is ready to merge. Note that this relies on an update to the QGIS project as well : https://github.com/QGEP/QGEP/pull/550 

Worth noting that on reach delete, the labels aren't updated, so I commented out this scenario from the tests (fixing this is better done separately, as it was already the case before the PR, and there may be side effects ?).

test/test_triggers.py
```diff
@@ -75,6 +75,145 @@ def test_identifier(self):
...
+            'rp_from_level': None
+        }
+        self.update('vw_qgep_reach', row_reach_c, reach_c_id)
+
+        expected.update({
+            '_output_label': '\nO1=?\nO2=92.00',
+        })
+        check_values(expected)
+        
+        # TODO : reenable this (currently, deleting reaches doesn't trigger update on labels)
```
m-kuhn: @olivierdalang is this an open todo you plan to work on?

olivierdalang: Probably better to do it in another PR, as it requires enabling ON DELETE triggers, which may have some side-effects. Is it OK to leave the TODO until then ?

m-kuhn: Yes, thanks!

PR #128, opened 2020-02-26
Docker setup for development

olivierdalang:
Not sure if this is useful to others, but this is the docker setup I made for my dev env.

I allows to set up postgis databases for:
- *release* : installs the demo data from the release 1.4
- *release_struct* : installs structure (empty model) from the release 1.4
- *build* : installs the structure using installation scripts
- *build_pum* : installs the demo data through successive pum migrations

It can be used both when developing the model (incl. to run tests) or as a test database for the QGIS plugin.

Oddly I had to remove the top level __init__.py for nose to run the tests. Is it used ? I don't think the root directory is supposed to be a python module, is it ?


m-kuhn:
Sweet, thanks for sharing

Can you move it to a subfolder (preferably `.docker`)?

@3nids can you have a look too?

olivierdalang:
If you think it's useful, ideally I'd refactor it so it all holds in one docker image and gets pushed to dockerhub, so that users would be able to spin up the datamodel with one line of command (obviously provided docker is installed).

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: "3"
+
+services:
+
+  postgis:
+    image: mdillon/postgis:9.6
```
m-kuhn:
```suggestion
    image: kartoza/postgis:11.0-2.5

```

Does that work too? Let's go a bit more modern ;)

m-kuhn:
Are you also thinking to auto deploy from CI?

olivierdalang:
Yes of course we could easily test against a matrix of postres/postgis versions, and yes auto-deploy for master and tags would be nice.

olivierdalang:
I merged everything in 1 container so it will be easier to run directly from docker hub, and will start tinkering on auto-builds and tests once it's merged.

Just waiting for https://github.com/opengisch/pirogue/pull/7 to be merged first though (for debian stretch's python 3.5)

.docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
+ARG POSTGIS_VERSION=9.6-2.5
```
m-kuhn: Why not use https://github.com/kartoza/docker-postgis/tree/11.0-2.5 instead? That should ship a newer python version too.

olivierdalang: I was used to mdillon, but saw it was moved to the postgis organization, so it's what's closest to an official release ?
But actually, I'm starting to think merging it in 1 container was not the best idea... basically it just saves 1 command line, and makes it much harder to swap the postgis image... Maybe we can change that at some point, so that we could use whatever postgis image/tag we want.

olivierdalang: But yes indeed it's based on buster, so python wouldn't have been an issue...

olivierdalang:
Ok I think this is **finally** ready to be squashed&merged.

The only point that may affect some usage is the deletion of the root level `__init__.py` file which prevents to run nosetests for some reason (not sure if it was serving any purpose though ?).

It's not 100% complete (not all tests are run), but it should already be useful as it is supposed to push the images to dockerhub.

PR #131, opened 2020-03-12
update catchment area

tproduit:
@sjib can you check if my modifications match what we disscussed during the code sprint?
https://github.com/QGEP/QGEP/issues/560
https://github.com/QGEP/QGEP/issues/561


ponceta:
@sjib @tproduit @m-kuhn merge-ready?

delta/delta_1.4.1_swmm_catchment_areas.sql
```diff
@@ -0,0 +1,231 @@
+DROP VIEW IF EXISTS qgep_swmm.vw_subcatchments CASCADE;
+CREATE OR REPLACE VIEW qgep_swmm.vw_subcatchments AS
```
m-kuhn, line 2: @3nids should all views be (re-created) in pre- / post-all?

delta/delta_1.4.1_swmm_catchment_areas.sql
```diff
@@ -0,0 +1,231 @@
...
+    st_maxdistance(st_centroid(perimeter_geometry), ST_ExteriorRing(perimeter_geometry))
+    + st_distance(st_centroid(perimeter_geometry), ST_ExteriorRing(perimeter_geometry))
+    )/2
+  END as Width, -- Width of overland flow path estimation
+  0.5 as percSlope, -- default value
+  0 as CurbLen, -- default value
+  NULL::varchar as SnowPack, -- default value
+  ca.identifier || ', ' || ca.remark as description,
+  ca.obj_id as tag,
+  ST_Simplify(ST_CurveToLine(perimeter_geometry), 5, TRUE) as geom
```
m-kuhn, line 31: Do we need them to be topologically correct still? Then ST_Simplify is likely not the right tool to use.

tproduit: I dont think that the topology play a role in SWMM. It is more the area which is important.

delta/delta_1.4.1_swmm_catchment_areas.sql
3nids: @m-kuhn yes. but this should be done in
https://github.com/QGEP/datamodel/blob/master/view/create_views.py#L74
https://github.com/QGEP/datamodel/blob/master/view/drop_views.sql

m-kuhn: CC @tproduit

delta/delta_1.4.1_swmm_catchment_areas.sql
sjib: Why don't we just take the topology as is? Any reason to simplify it?

tproduit: I don't remember why I simplified the geometry. There is probably no reason except to have a more schematic map in SWMM GUI. I will remove the simplification

delta/delta_1.4.1_swmm_catchment_areas.sql
tproduit: I have added the swmm view in these two files

PR #133, opened 2020-04-03
Network rewamp

olivierdalang:
Rewrite of the network generation logic to fix issues with blind connections and prepare for future profile tool enhancement, as discussed during hackfest. 

This requires an update to qgep plugin https://github.com/QGEP/qgepplugin/pull/51

This fixes https://github.com/QGEP/QGEP/issues/556

Here are the differences :

- Graph nodes are now created for blind connections, so that the graph is now a proper graph and should yield more reliable results.
- To simplify the SQL,  the network is stored in tables with an update function instead of using materialized views (meaning the `REFRESH MATERIALIZED VIEW qgep_od.vw_network_*;` queries must be replaced by `SELECT qgep_network.refresh_network_simple();`).
- These tables only contain pure network information with reference to ids. It is intended to be used only for graph related queries. The future profile tool will retrieve features attributes from the qgep layer directly, so that we avoid duplicated and out of sync data (e.g. you wouldn't need to rebuild the graph to get updated levels on the profiles). 
- Backwards compatible materialized views are provided (and refreshed by the above function), so that the profile tool should continue to work exactly in the same way for now. This would be removed once we have a new profile tool.


m-kuhn:
Really good work !
Thanks for all the tests.

@urskaufmann could you execute the new network tracing sql on your database and check if the results are reliable?

https://github.com/QGEP/datamodel/blob/7c5d6c00db61597c6c86a8ee560ec0c7db5a561e/delta/delta_1.5.0_fix_network_tracking_views.sql

urskaufmann:
I have now a schema qgep_network in my database, but both tables are empty.
I used the SQL-Button in the QGEP-project. And now also the vw_network_node and -segment are both empty in schema qgep_od.
?

m-kuhn:
@urskaufmann can you run `SELECT qgep_network.refresh_network_simple()` once? The refresh network button will in the future do that also but not yet.

urskaufmann:
Yeaaah!!!
Seems to work great.

PR #135, opened 2020-05-05
Fix qgep_od.vw_file unnecessary prefix for file kind

ponceta:
To avoid typo errors and redundancy.

### DIFF
--- a/view/vw_file.sql
+++ b/view/vw_file.sql
@@ -18,7 +18,7 @@
 CREATE OR REPLACE VIEW qgep_od.vw_file AS
   SELECT f.obj_id,
     f.identifier,
-    f.kind AS file_kind,
+    f.kind,
     f.object,
     f.class,
     -- dm.path,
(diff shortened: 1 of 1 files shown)


ponceta:
SELECT obj_id, url
FROM (
  SELECT ws.obj_id AS obj_id,
  file._url as url,
  rank() OVER (PARTITION BY ws.obj_id ORDER BY file.obj_id ASC) as file_index
  FROM qgep_od.vw_qgep_wastewater_structure ws 
  LEFT JOIN qgep_od.vw_file file ON ws.obj_id = file.object
  **WHERE file.file_kind = 3772 -- Filter only images** -- Boring syntax
) AS files
WHERE file_index = 2 -- Index starts at 1

3nids:
what do you mean by typo error and redundancy?
There is a `kind` field in another table?

If we go through with this, the demo project should be adapted too.

PR #136, opened 2020-06-04
Fix failing tests due to network rewamp

olivierdalang:
(deletes didn't cascade to the network table)

This PR also fixed the Docker CI setup so that all tests are now run on github actions as well. This means tests should pass/fail identically on both travis and github workflows.


m-kuhn:
Thanks a lot !

PR #137, opened 2020-06-04
Run vacuum on network tests

m-kuhn:
Hopefully this fixes travis tests running out of space

https://travis-ci.org/github/QGEP/datamodel/builds/694673273#L700

### DIFF
--- a/test/test_network.py
+++ b/test/test_network.py
@@ -25,6 +25,7 @@ class TestNetwork(unittest.TestCase, DbTestBase):
 
     def tearDown(self):
         self.conn.rollback()
+        self.vacuum()
 
     def setUp(self):
         pgservice=os.environ.get('PGSERVICE') or 'pg_qgep'
--- a/test/utils.py
+++ b/test/utils.py
@@ -117,3 +117,11 @@ class DbTestBase:
         Helper to make 2D point geometries
         """
         return self.execute('ST_SetSrid(ST_MakePoint(%s, %s), %s)', [x, y, srid])
+
+    def vacuum(self):
+        old_isolation_level = self.conn.isolation_level
+        self.conn.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
+        query = "VACUUM FULL"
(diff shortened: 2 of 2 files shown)


m-kuhn:
@olivierdalang let's hope this improves things.

FYI, I also changed line endings to unix ones, that's why the diff is huge. Here's a relevant one:
https://github.com/QGEP/datamodel/pull/137/files?diff=split&w=1

m-kuhn:
@olivierdalang 
If I get that correctly, all tests succeed on gh workflows. We also have a docker image pushed from gh workflows.
I wonder if instead of fixing the tests on travis we should rather just move the missing bits to the workflows and disable travis.

Currently, only the only thing that comes to my mind is "publish dumps to release", do you agree?

olivierdalang:
@m-kuhn 
Yes, adding the publishing of dumps should be easy to add, and as all tests are now run on GH workflows, testing is now redundant (at least theoretically)...

I'm wondering however if the full disk failure doesn't hide anything more serious, as the test dataset isn't really big ? (not too familiar with travis, but I feel other projects would use much more disk quota than our tests ?)

m-kuhn:
My main suspicion is that it's caused by all the rollbacks (that's what the tests rely on for isolation).
This will revert logical changes, but keep physical changes as far as I understand. I don't exactly understand when it's freed again, I was hoping for vacuum to clean up, but maybe it needs something more like the connection being closed.

m-kuhn:
TODO:

- [ ] Locally check the disk space behavior of the network tracking tables / creation
- [ ] Either: Fix if there is something to fix
- [ ] Or: Deploy via github workflows

olivierdalang:
I checked disk usage (through Docker) :
- creating the container is around 1.011 gb (but almost none of this happens on travis)
- initializing the databases (empty and from the dumps) adds 373 mb (that happens on travis too, but only after the failure, for the import_tests)
- disk usage only increases by around 1mb during the tests

So I'm rather clueless at what it could be... I'll try to fiddle a bit with the tests (change the order or so) to see if it always fails at the same place.

PR #140, opened 2020-06-22
Use ST_LineSubstring instead of legacy ST_Line_Substring

m-kuhn:
Fixes https://github.com/QGEP/QGEP/issues/577

### DIFF
--- a/.github/workflows/docker-test-and-push.yaml
+++ b/.github/workflows/docker-test-and-push.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         # postgres-postgis version, see available tags https://hub.docker.com/r/postgis/postgis/tags
-        pgis: ["9.6-2.5", "10-2.5", "11-2.5", "12-2.5"]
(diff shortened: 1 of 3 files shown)


olivierdalang:
In the diff, you just need the `CREATE OR REPLACE FUNCTION` part, the table and schema are there already (probably explains the failure)

Also I added a PR to test under postgis 3 here : https://github.com/QGEP/datamodel/pull/141
This would have catched the issue

m-kuhn:
> In the diff, you just need the `CREATE OR REPLACE FUNCTION` part, the table and schema are there already (probably explains the failure)

Can we get more verbose output in case of errors?

> 
> Also I added a PR to test under postgis 3 here : #141
> This would have catched the issue

:+1: nice one

olivierdalang:
> Can we get more verbose output in case of errors?

It's in the "failure logs" step (if you're talking about github workflows) :
```
...
     Applying delta 1.5.1 fix_network_tracking_views... schema "qgep_network" already exists
```

PR #141, opened 2020-06-22
run tests also against postgis 3 (in github ci)

olivierdalang:


olivierdalang:
failure under postgis 3 is expected and should be fixed by https://github.com/QGEP/datamodel/pull/140

m-kuhn:
Merged into https://github.com/QGEP/datamodel/pull/140

PR #142, opened 2020-06-22
Fix geometry types for swmm views

m-kuhn:
This allows for an easier import in qgis and more type safety

### DIFF
--- a/scripts/db_setup.sh
+++ b/scripts/db_setup.sh
@@ -93,23 +93,3 @@ pum baseline -p ${PGSERVICE} -t qgep_sys.pum_info -d ${DIR}/delta/ -b ${VERSION}
 
 
 psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/14_geometry_functions.sql
-
-## SWMM Views
-
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/01_vw_swmm_create_schema.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/02_vw_swmm_junctions.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/03_vw_swmm_aquifers.sql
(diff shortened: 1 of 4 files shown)


scripts/db_setup.sh
```diff
@@ -93,23 +93,3 @@ pum baseline -p ${PGSERVICE} -t qgep_sys.pum_info -d ${DIR}/delta/ -b ${VERSION}
...
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/08_vw_swmm_outfalls.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/09_vw_swmm_subcatchments.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/10_vw_swmm_vertices.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/11_vw_swmm_pumps.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/12_vw_swmm_polygons.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/13_vw_swmm_storages.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/14_vw_swmm_xsections.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/15_vw_swmm_coordinates.sql
-psql "service=${PGSERVICE}" -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/swmm_views/16_vw_swmm_tags.sql
-
```
olivierdalang, line 115: Not sure I understand why this is removed, but I guess they are already run from elsewhere ?

m-kuhn: It will be done in [create_views](https://github.com/QGEP/datamodel/blob/e31f44cf038d6b8f6bb608f038d1cf8023886d08/view/create_views.py) called here https://github.com/QGEP/datamodel/blob/e31f44cf038d6b8f6bb608f038d1cf8023886d08/scripts/db_setup.sh#L84 so was redundant and the style for SRID was incompatible.

PR #143, opened 2020-06-30
Swmm tests and fix

olivierdalang:


.github/workflows/docker-test-and-push.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: "run tests on qgep_build (structure)"
         run: |
           docker start qgep
-          docker exec -e PGSERVICE=qgep_build qgep pytest --ignore test/test_import.py
+          docker exec -e PGSERVICE=qgep_build qgep pytest --ignore test/test_import.py --ignore test/test_swmm.py
```
m-kuhn, line 46: Why do we ignore this test?

olivierdalang: the new test runs in the `run tests on qgep_build_pum (production)` step just after that one, as it relies on the demo data

m-kuhn:
```suggestion
          # Run tests that depend on data model. Tests that require demo data are ignored and run below in qgep_build_pum
          docker exec -e PGSERVICE=qgep_build qgep pytest --ignore test/test_import.py --ignore test/test_swmm.py
```

Can we add that for clarity?

olivierdalang: done, added the same to the travis file too

PR #146, opened 2020-07-29
fix for https://github.com/QGEP/QGEP/issues/591 : run drop_views.sql …

olivierdalang:
Fix for https://github.com/QGEP/QGEP/issues/591

This includes `drop_views.sql` in `pre-all`, hopefully solving a regular where migration would fail because you cannot alter an element used in a view. All views are recreated afterwards in post-all.

The same approach should be used not only for view, but for all "thin" objects (functions, triggers, etc.), that don't hold data, and as such that we don't need to actually migrate.
 
This PR does:
- [x] A. adds `drop_view` in pre-all.py
- [x] B. adds a version check on PUM and PIROGUE in pre-all.py (older version of pum don't handle changes in existing delta files)
- [x] C. removing all references to objects in migrations that are dropped/recreated by `drop_view.sql` and `create_views.py` views in migrations, and that now fail because the views are dropped pre-migration 
- [x] D. move triggers on qgep_import.manhole_quarantine from `13_import.sql` to `views/vw_qgep_import.sql` (not sure why this started to be an issue, maybe removed too much stuff in C)

Note that there are still quite a few views/triggers/functions that are not yet in the drop/create scripts, and that would benefit from being there as well. I think we can improve this slowly over time (each time that we need to alter these objects, instead of creating a migration, we could move them to pre/post and remove references in the migrations).

⚠️ WARNING obviously this is quite a big change, and touching existing migrations files is quite risky, so please review this really carefully ⚠️


olivierdalang:
Yes, so it means we'd have to remove all references to views in old migrations files.

This can't be done before https://github.com/opengisch/pum/pull/81 is merged and a new pum is releases, as by changing the deltas, pum may try to reapply them (as their checksum will change).

olivierdalang:
@3nids I guess you're the most aware of whether this (removing views from the deltas/datamodel.sql scripts and always forcing drop_views beforehand) is a sound approach ?

3nids:
> removing views from the deltas/datamodel.sql scripts and always forcing drop_views beforehand

yes, that's the way to go. You can look in qwat/qwat-data-model, it's done like this.

ponceta:
@olivierdalang to run this we have to remove references to views in the previous delta (was already made on QWAT a few years ago)

olivierdalang:
If we implement that, **we need to include a check on the pum version**, probably in pre-all.py, and fail if the installed version of pum doesn't include https://github.com/opengisch/pum/pull/81, otherwise this will likely cause a lot of trouble

sjib:
Just supporting the latest pum version and forcing people to go for that I think is a good idea anyway - or are there any reason to not do that?

olivierdalang:
I think there's no reason not to do it, mostly added that comment to not forget it :-)

ponceta:
This will drop much actual delta complexity, and it is not easy for the users to install and to know which old pum version they install (since the doc only provide `pip install pum` )

ponceta:
@olivierdalang this looks very good work, thank you, I'll run some update tests on real database too.

@varrieta could you have a look at this too? Especially the part on the import.sql alterations. (It sould work the same as before)

@sjib this should solve your issue with some views in the deltas.

delta/delta_1.4.0_swmm_views.sql
```diff
@@ -1,721 +1 @@
-DROP SCHEMA IF EXISTS qgep_swmm CASCADE;
```
ponceta, line 1: This is not mandatory to remove, but I see no objection to remove it

ponceta (approved):
Except a small comment, I did not find any bugs. And this is mainly a reorganisation of the code so it needs testing in priority.

I continue testing on real data to give some more feedback.

ponceta:
Delta delta_1.5.5_add_spatial_index_on_network_views.sql is not passing on production database and **should make the CI fail** :

Content is :

```
CREATE INDEX in_qgep_od_vw_network_segment_progression_geometry ON qgep_od.vw_network_segment USING gist (progression_geometry);
CREATE INDEX in_qgep_od_vw_network_node_situation_geometry ON qgep_od.vw_network_node USING gist (situation_geometry);
```

ponceta:
@olivierdalang I've proposed an additionnal commit to address the network views and test it in a separate PR, could you have a look at it?

https://github.com/QGEP/datamodel/pull/205

delta/delta_1.4.0_swmm_views.sql
olivierdalang: Yes, it wasn't totally clear for me what we should do with these statements (also left in quite a few `DROP ... IF EXISTS` that we could get rid of theoretically since now done in pre-all...

olivierdalang:
About dropping/recreating the materialized view, I'm not 100% sure, as there is a much higher cost than a regular view (since it will also drop all data). I'm not sure how much time recreating the network views takes on large dataset, but that would become something that must be done for each upgrade (which may become much more frequent in a rolling release model).

But I haven't seen the issue itself, can you copy the error message we get without your fix ? Ideally as you said, the CI should fail like your production data. Do you know why it doesn't ?

I'd suggest first we make CI fail as it does in prod, then we see how we can fix it.

olivierdalang:
Ok included your commit. Ok to merge ? (since not yet in rolling release model, this can still be tested)

PR #147, opened 2020-08-18
gruner improvements

tproduit:
Improvements and corrections of the swmm views after @DflGruBoe tests.


tproduit:
@m-kuhn in this merge request, I did some minor changes to the swmm views according to the feedback of Grüner. Is there a way to have more details on what makes the checks fail?

ponceta:
Since views should be dropped before and restored after pum update, there is no need to create delta files but actual views should be modified! That's probably what is failing here.

m-kuhn:
@olivierdalang could you give this a look?

olivierdalang:
@tproduit 
>Is there a way to have more details on what makes the checks fail?

Yes you can see the full log here : https://github.com/QGEP/datamodel/runs/996972283 (you get that link by clicking on the ❌ , then on details, for most failures it doesn't matter which version's logs you look at as the failure will be the same).
You then see a list of steps that have been run for automatic testing, and in case of failure, one additional `failure logs` stepis run. Unfold this, and scroll to the very bottom to see the full errors.

In the case of this PR, it is :
```
[...]
     Applying delta 1.4.0 geometry_not_null_constraints... OK
     Applying delta 1.4.0 import_views... OK
     Applying delta 1.4.0 increase_maintenance_event_base_data_length... OK
     Applying delta 1.4.0 municipality_precision... OK
     Applying delta 1.4.0 swmm_views... OK
     Applying delta 1.4.0 update_wastewater_structure_label_with_bottom_level... OK
     Applying delta 1.4.1 swmm_catchment_areas... OK
     Applying delta 1.5.0 fix_network_tracking_views... OK
     Applying delta 1.5.1 fix_network_tracking_views... OK
     Applying delta 1.5.2 fix_network_tracking_views... OK
     Applying delta 1.5.3 fix_vl_value_en_digital_video... OK
     Applying delta 1.5.4 swmm_views_improvements... cannot drop view qgep_swmm.vw_conduits because other objects depend on it
DETAIL:  view qgep_swmm.vw_vertices depends on view qgep_swmm.vw_conduits
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

I think we need to clarify how we want to deal with views (here : https://github.com/QGEP/datamodel/pull/146), as currently, it is a bit ambiguous.

tproduit:
@olivierdalang if I understand well #146  , I have to remove my delta file as the view will be dropped and recreated?

olivierdalang:
@tproduit 
Yes exactly, no need to drop/create view in migrations, as all views will be recreated after all other migrations finished (due to `post-all.py` in the delta directory)

ponceta:
LGTM!

test/test_swmm.py
```diff
@@ -81,7 +81,7 @@ def test_count_vw_vertices(self):
         self.assert_count("vw_vertices", "qgep_swmm", 32)
 
     def test_count_vw_xsections(self):
-        self.assert_count("vw_xsections", "qgep_swmm", 2348)
+        self.assert_count("vw_xsections", "qgep_swmm", 17200)
```
olivierdalang, line 84: Is this change of result expected from the above changes ?

From the code, it seems to indicate that `ws._function_hierarchic` is not the same as `ch.function_hierarchic`, which seems surprising, no ? Could it indicate some other issue, e.g. a trigger that doesn't properly update it ?

tproduit: Indeed if you run: `
SELECT DISTINCT
  re.obj_id as Link,
  ws._function_hierarchic,
  ch.function_hierarchic
FROM qgep_od.reach re
LEFT JOIN qgep_od.pipe_profile pp on pp.obj_id = re.fk_pipe_profile
LEFT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id::text = re.obj_id::text
LEFT JOIN qgep_od.wastewater_structure ws ON ws.obj_id::text = ne.fk_wastewater_structure::text
LEFT JOIN qgep_od.channel ch ON ch.obj_id::text = ws.obj_id::text
WHERE ch.function_hierarchic = ANY (ARRAY[5066, 5068, 5069, 5070, 5064, 5071, 5062, 5072, 5074]);
`
The two columns are not similar. I think that the trigger that computed ws._function_hierarchic was mainly thought for the point-like objects (manhole etc.)

urskaufmann: > The two columns are not similar. I think that the trigger that computed ws._function_hierarchic was mainly thought for the point-like objects (manhole etc.)

That's right! The two columns can be different.

tproduit: So yes the change of result is expected from the above change.

PR #152, opened 2020-09-01
fix output label ordering

3nids:
with a test 💪

### DIFF
--- a/.docker/init_qgep.sh
+++ b/.docker/init_qgep.sh
@@ -22,9 +22,9 @@ recreate_db(){
 }
 
 if [ "$1" == "wait" ]; then
-
+  printf "initializing QGEP…"
   until [ -f ${PGDATA}/entrypoint-done-flag ]; do
-    printf "🐘"
+    printf " 🐘"
     sleep 3
   done
   echo ""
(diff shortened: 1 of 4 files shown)


3nids:
@olivierdalang this is now ready for review :)

olivierdalang:
The fix looks good to me.

As far as I understood the bug, it appears for reaches with strong curvature (where the azimuth incorrectly calculated for outputs at ST_LineInterpolatePoint(..., 0.99) vs the one correctly calculated at ST_LineInterpolatePoint(..., 0.01) may be different), right ?
If so, I'm not sure the test checks for that case ? (as it seems all reaches have just two points, one being at the main manhole, so that the azimuth should be the same at any point on the line). Does it fail without the fix ?

Less importantly, there was already very basic testing of label generations in https://github.com/QGEP/datamodel/blob/master/test/test_triggers.py#L182-L192, maybe we could put all related tests in the same place ?

3nids:
> As far as I understood the bug, it appears for reaches with strong curvature (where the azimuth incorrectly calculated for outputs at ST_LineInterpolatePoint(..., 0.99) vs the one correctly calculated at ST_LineInterpolatePoint(..., 0.01) may be different), right ?

Indeed, I realized it. I have updated the tests.

kandre:
Thanks @3nids, looks good to me.

PR #154, opened 2020-09-18
Channels in vw_qgep_wastewater_structure

3nids:
fix https://github.com/QGEP/QGEP/issues/405

### DIFF
--- a/view/vw_qgep_wastewater_structure.py
+++ b/view/vw_qgep_wastewater_structure.py
@@ -85,8 +85,8 @@ def vw_qgep_wastewater_structure(srid: int,
         LEFT JOIN qgep_od.special_structure ss ON ss.obj_id = ws.obj_id
         LEFT JOIN qgep_od.discharge_point dp ON dp.obj_id = ws.obj_id
         LEFT JOIN qgep_od.infiltration_installation ii ON ii.obj_id = ws.obj_id
-        LEFT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
-        LEFT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
+        RIGHT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
+        RIGHT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
         {extra_joins};
 
         ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_structure');
--- a/view/vw_qgep_wastewater_structure.py
+++ b/view/vw_qgep_wastewater_structure.py
@@ -89,7 +89,7 @@ def vw_qgep_wastewater_structure(srid: int,
         LEFT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
         LEFT JOIN qgep_od.channel ch ON ch.obj_id = ws.obj_id
         {extra_joins}
-        WHERE ch.obj_id IS NOT NULL;
+        WHERE ch.obj_id IS NULL;
 
         ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','wastewater_structure');
         ALTER VIEW qgep_od.vw_qgep_wastewater_structure ALTER co_obj_id SET DEFAULT qgep_sys.generate_oid('qgep_od','cover');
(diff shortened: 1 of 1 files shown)


m-kuhn:
I think this will hide structures without a node (cover only)

view/vw_qgep_wastewater_structure.py
```diff
@@ -85,8 +85,8 @@ def vw_qgep_wastewater_structure(srid: int,
         LEFT JOIN qgep_od.special_structure ss ON ss.obj_id = ws.obj_id
         LEFT JOIN qgep_od.discharge_point dp ON dp.obj_id = ws.obj_id
         LEFT JOIN qgep_od.infiltration_installation ii ON ii.obj_id = ws.obj_id
-        LEFT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
-        LEFT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
+        RIGHT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
+        RIGHT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
         {extra_joins};
```
m-kuhn, line 90:
```suggestion
        LEFT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
        LEFT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
        LEFT JOIN qgep_od.channel ch ON ch.obj_id = ws.obj_id
        {extra_joins}
        WHERE ch.obj_id IS NOT NULL;
```

m-kuhn: Counter proposal

sjib: the view should contain all elements of qgep_od_wastewater_structure except those who are also part of qgep.od_channel

sjib: "Could this just be filtered by ws_type?" - This is tricky as this could be in contradiction to whether an wastewater_structure is really part of the respective subclass. I would not go on this. In the demodata most channels have currently ws_type = "unknown". ws_type is a additional internal attribute that was set up to filter the data needed on the input form.

m-kuhn: Correction: ws_type is not stored, **it's derived from a check on which of the subtables the object is actually found**.

The current counter proposal also checks for existence on the subtable channel as well.

sjib: So what does it mean if ws_type is "unknown" in a project?

m-kuhn: At the moment: only a superclass entry and no entry in any of the subclass tables. Probably channel.

```
    CASE
      WHEN ma.obj_id IS NOT NULL THEN 'manhole'
      WHEN ss.obj_id IS NOT NULL THEN 'special_structure'
      WHEN dp.obj_id IS NOT NULL THEN 'discharge_point'
      WHEN ii.obj_id IS NOT NULL THEN 'infiltration_installation'
      ELSE 'unknown'
    END AS ws_type,

```

sjib: Then it would make sense to extend this statement to inlcude all existing subclasses, so that we only get unknown if data is not added correctly (not an entry in both sub and superclass):

CASE
      WHEN ma.obj_id IS NOT NULL THEN 'manhole'
      WHEN ss.obj_id IS NOT NULL THEN 'special_structure'
      WHEN dp.obj_id IS NOT NULL THEN 'discharge_point'
      WHEN ii.obj_id IS NOT NULL THEN 'infiltration_installation'
      WHEN cl.obj_id IS NOT NULL THEN 'channel'
      WHEN  wt.obj_id IS NOT NULL THEN 'wwtp_structure'
      ELSE 'unknown'
    END AS ws_type,

m-kuhn: https://github.com/QGEP/QGEP/issues/604

ponceta: In Pully we have currently:
`WHERE ws.fk_main_wastewater_node IS NOT NULL;`

view/vw_qgep_wastewater_structure.py
```diff
@@ -87,7 +87,9 @@ def vw_qgep_wastewater_structure(srid: int,
         LEFT JOIN qgep_od.infiltration_installation ii ON ii.obj_id = ws.obj_id
         LEFT JOIN qgep_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
         LEFT JOIN qgep_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
-        {extra_joins};
+        LEFT JOIN qgep_od.channel ch ON ch.obj_id = ws.obj_id
+        {extra_joins}
+        WHERE ch.obj_id IS NOT NULL;
```
3nids, line 92: am I mixing things, or shouldn't it be the `IS NULL`?

m-kuhn:
```suggestion
        WHERE ch.obj_id IS NULL;
```

correct, thanks

3nids (commented):
I cannot self-approve on github, but I approve the changes!

m-kuhn (approved):

ponceta:
Thanks for the fix!

PR #160, opened 2020-10-28
adds a CI check to assert CURRENT_VERSION.txt is up to date

olivierdalang:
(which seemed not to be the case between versions 1.5.1 and 1.5.2, not sure it has impact on releases but still...)


.github/workflows/docker-test-and-push.yaml
```diff
@@ -1,92 +1,101 @@
...
+      - uses: actions/checkout@v2
+
+      - name: "assert version is up to date"
+        run: |
+          python -c '
+          import os
+          cur_version = open("system/CURRENT_VERSION.txt","r").read().strip()
+          last_delta_version = sorted(list([d for d in os.listdir("delta") if d.startswith("delta_")]))[-1].split("_")[1]
+          assert cur_version == last_delta_version, cur_version + " != " + last_delta_version
+          '
```
3nids, line 38:
```suggestion
          LAST_VERSION=$(ls delta | sort -r | sed '/^delta_/!d' | head -n 1 | cut -d_ -f2)
          CUR_VERSION=$(cat system/CURRENT_VERSION.txt)
          [[ "${LAST_VERSION}" != "${CUR_VERSION}" ]] && echo "Version mismatch ${LAST_VERSION} != ${CUR_VERSION}" && exit 1
```

I tend to prefer native scripting ;)

also, are the white spaces removal intentional in the rest of the file?

olivierdalang: Ok made the change and reverted unintentional whitespace

olivierdalang:
Doesn't work, should it be 
`[[ "${LAST_VERSION}" == "${CUR_VERSION}" ]] || echo "Version mismatch ${LAST_VERSION} != ${CUR_VERSION}" && exit 1`
instead of
`[[ "${LAST_VERSION}" != "${CUR_VERSION}" ]] && echo "Version mismatch ${LAST_VERSION} != ${CUR_VERSION}" && exit 1` ?

(not too expert with bash)

PR #162, opened 2020-11-19
crlf -> lf on whole codebase

olivierdalang:
Seems we have some crlf ending files, which started causing issues here (TRIGGER functions considered as different because of different line endings in their definition).

I'm not exactly sure if this commit fixes it, as I have some strange things happening with my git, maybe someone with a proper os could run os2unix on the whole codebase ?

Also before merging let's make sure we don't have pending PRs that we want to merge to avoid conflicts.


3nids (approved):

m-kuhn:
What about a sustainable solution?
image

m-kuhn:
```
echo "* text=auto" > .gitattributes
git add --renormalize .
```

olivierdalang:
(superseeded by https://github.com/QGEP/datamodel/pull/165)

PR #163, opened 2020-11-24
Swmm states

tproduit:
Add states in the view and solve several errors and bug.

     

-    Only primary network is exported
-    Most objects have a state attribute: "planned" or "current". This state allows the user to create two types of export "planned" or "current".
-    Blind connections are exported (primary reach connected without a manhole)

Fixes QGEP/QGEP#592
Fixes QGEP/QGEP#561
Fixes QGEP/QGEP#559


tproduit:
Fixes https://github.com/QGEP/QGEP/issues/592
Fixes https://github.com/QGEP/QGEP/issues/561
Fixes https://github.com/QGEP/QGEP/issues/559

olivierdalang (commented):
Hi !

Just had a look at the code, technically it makes sense, I did not see anything that could cause issues. 

I see the row counts changed in the tests. If it was just about adding the "state" column being added, I'd expect the numbers to be the same (or at least that they all get smaller if we don't export some states anymore). Are the new numbers expected ? (it's a bit hard to tell as some other business-logic changes are also present in the PR). These tests are a bit crude, but still the only way we have so far to detect unexpected side effects.

Also, besides the docs, it would be great to have a short high-level descriptions about the changes meant for users, that we can reuse when we do a release, that explains the new feature and the breakages (if any) for current workflows (eg. what objects are now excluded from exports).

tproduit:
> I see the row counts changed in the tests. If it was just about adding the "state" column being added, I'd expect the numbers to be the same (or at least that they all get smaller if we don't export some states anymore). Are the new numbers expected ? (it's a bit hard to tell as some other business-logic changes are also present in the PR). These tests are a bit crude, but still the only way we have so far to detect unexpected side effects.

The new number are expected. Some objects in some views might yet appear several times. I'm also not 100% convinced by these tests. It is likely that any consequent change in the views will change also the number of objects.

tproduit:
> Also, besides the docs, it would be great to have a short high-level descriptions about the changes meant for users, that we can reuse when we do a release, that explains the new feature and the breakages (if any) for current workflows (eg. what objects are now excluded from exports).
Main changes:
* Only primary network is exported
* Most objects have a state attribute: "planned" or "current". This state allows the user to create two types of export "planned" or "current".
* Blind connections are exported (primary reach connected without a manhole)

PR #164, opened 2020-12-07
Role permissions: separate cluster initialisation from db initialisation

m-kuhn:
Untangles

 - role creation - one shot cluster initialization jobs
 - permissions - that need to be granted once per database

### DIFF
--- a/12_1_roles.sql
+++ b/12_1_roles.sql
@@ -1,15 +1,6 @@
-/* create roles */
-DO $$
-DECLARE
-    role text;
-BEGIN
-    FOREACH role IN ARRAY ARRAY['qgep_viewer', 'qgep_user', 'qgep_manager', 'qgep_sysadmin'] LOOP
-      IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = role) THEN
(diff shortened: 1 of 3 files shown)


12_1_roles.sql
```diff
@@ -1,15 +1,6 @@
...
-      END IF;
-    END LOOP;
-END
-$$;
+------------------------------------------
+/* GRANT on schemas - once per database */
+------------------------------------------
 
 /* Viewer */
 GRANT USAGE ON SCHEMA qgep_od  TO qgep_viewer;
```
m-kuhn, line 6: @ponceta are the permissions on the sequences required for a viewer or can that be granted for user only?

3nids (approved):

12_1_roles.sql
sjib: Everything what is needed to run the calculations for labelling or additional attributes such as _*functionhierarchic has to be granted, else we will have wrong labellings in the view I guess.

m-kuhn: Do we need sequences for this?

ponceta: To my knowledge it is not required, but I rarely use the qgep_viewer role.

PR #166, opened 2021-01-13
Fix 350 (usage current and function hierarchic on nodes)

olivierdalang:
This PR implements https://github.com/QGEP/QGEP/issues/530.

I adapted the function that updates these columns on wastewater_structures.

For now, it's just a first implementation for initial review on the approach. Left to do :

- [x] (as per comment in code) Check whether the logic in the pre-existing trigger is actually correct. The issue is both fk_reach_point_from and fk_reach_point_to are used to determine which wastewater structure/nodes to update [1], while the script uses `SELECT INTO` [2], which only supports one row. IMO this can not work as expected, as it would update just one of the two node/structure of the reach, resulting in non-reliable updating by the trigger. Is it actually working correctly currently ? @m-kuhn what do you think ? (I think you wrote that, but several years ago :-) )

- [x] ~~Currently `vw_wastewater_node` is not styled. Do we want to update the QGIS project to reflect this ? How exactly ?  At least we could colorize according to `_usage_current`. @ponceta Do you have a suggestion for this ?~~

- [x] Tests would probably be welcome

[1] : see https://github.com/QGEP/datamodel/blob/master/06_symbology_functions.sql#L66
[2] : see https://github.com/QGEP/datamodel/blob/master/06_symbology_functions.sql#L63


ponceta:
I'll run a few tests on this tomorrow and I come back to you!

olivierdalang:
I wrote a very basic test that covers the issue mentioned above, and included a fix for it. It would be nice to have a careful review though as I may be missing something (especially if everything was working as expected before this PR despite the bug I think I've found :-) ).

ponceta:
I ran the 1.5.5 delta on our actual datamodel and when running 

`SELECT qgep_od.update_wastewater_node_symbology(NULL,true);`

It runs forever, 

image

ponceta:
Otherwise it is very functionnal on new entities. 

image

image

olivierdalang:
> I ran the 1.5.5 delta on our actual datamodel and when running
> 
> `SELECT qgep_od.update_wastewater_node_symbology(NULL,true);`
> 
> It runs forever,

Indeed... I don't understand exactly what's happening. It seems some rows result in an infinite trigger loop. In the mean time, I you disable the triggers before updating running the function on all rows it seems to work.

```
SELECT qgep_sys.drop_symbology_triggers();
SELECT qgep_od.update_wastewater_node_symbology(NULL, true);
SELECT qgep_sys.create_symbology_triggers();
```

Still need to find out what exactly is going on...

@ponceta About the symbology, do you have a suggestion, or should we leave

ponceta:
@olivierdalang thanks for the info, indeed it makes the job when the trigger is down :

image

For the symbology I would keep it that way for the moment, demo data does not includes lonely wastewater nodes so it's difficult to make it relevant.

PR #169, opened 2021-01-15
[wip] fix functions inconsistencies

olivierdalang:
Hi !

It seems there are some inconsistencies in functions between the migrated demo data and a freshly created database.

Upon inspection, it seems that paramter-less functions were overlooked by pum check (see https://github.com/opengisch/pum/pull/97).

This can be quite error prone, e.g. for the obsolete qgep_od.drop_symbology_triggers(), which clashes with qgep_sys.drop_symbology_triggers(), and/or introduce nasty bugs because of differences in functions btw updated demo data vs freshly created database.

This PR pushes pum version (TODO : release new pum and use that instead of temporary archive), and adds a delta that tries to fix the situation. 

I'm not exactly sure if that's just the case for the demo data, or if there's a chance that actual prod database have similar issues, but I guess there's a relatively high chance that this will make `pum check` report errors where before it was happy, meaning upcomming trouble on upcomming datamodel updates.

The delta to fix inconsistencies is not complete yet (hence failing CI), but please let me know if you think the approach makes sense.


olivierdalang:
This is almost ready, but there's one question left

- Some tables have different column orders between released datamodel and migrated datamodel. This lead to different column order in views, and thus in triggers functions created by pirogue.
- Thus, this PR relies on an update to pirogue that sorts columns in views/triggers functions, to allow for exact equality comparison 
- An arguably cleaner approach, would be to have `pum check` check for columns order too, so that we ensure released datamodel and migrated datamodel have same columns order, as this could have some other subtle side effects. The main drawback with that is that postgres doesn't allow changing column order, meaning that we'd be condemned to always keep the same column order in the schema definition (we could only add new columns at the end).

What do you think is best ?

If we're the current approach of this PR, I'll just make a new release of pirogue to pypi before merging this.

3nids:
I am -1 on checking for column orders in tables since this will prevent us from adding new columns at their natural position (only at the end as you mentionned).
I am +0 on the change done to pirogue, for the same reason, but in the end who cares about the order in a trigger.
I am really happy with how you solved this.

3nids (approved):

ponceta:
@olivierdalang I join @3nids on the comment about columns order, but as far as I'm concerned, if the user is able to disable the check on columns order, it's fine for me.

Let me know it you think there's a good motivation to merge this (more benefits than trouble in future updates)

olivierdalang:
@ponceta Hard to tell about whether it will bring more benefits or trouble. For sure it could detect inconsistencies that for now are silently ignored (which depending on the severity of the inconsistency could either be harmless errors preventing simple upgrades, or saving user from potentially serious bugs).

olivierdalang:
(superseded by https://github.com/QGEP/datamodel/pull/184)

PR #170, opened 2021-02-04
Roles fix for network

olivierdalang:
This PR adds some fixes to allow qgep users (qgep_viewer, qgep_user, ...) to correctly use the network schema.

- `qgep_viewer` can now properly select in the qgep_network schema (probably not too useful for now as we mostly use the compatibility views in qgep_od, but it doesn't hurt).

- `qgep_user` has now all permissions on tables in qgep_network (again not too useful, as normally he doesn't need to manually interact with these tables).

- `qgep_network.refresh_network_simple()` now uses `SECURITY DEFINER` so that it will work for any user who executes it (as it requires to truncate the network tables).

It looks like in postgres, the execute function permission is there by default (at least it seems qgep_viewer is able to call it, even without setting that permission explicitly). Is that correct ? This means qgep_viewer can refresh the network tables, but I think that's even better.


07_views_for_network_tracking.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE qgep_network.segment (
   geom geometry('LINESTRING', :SRID)
 );
 
-CREATE OR REPLACE FUNCTION qgep_network.refresh_network_simple() RETURNS void AS $body$
+CREATE OR REPLACE FUNCTION qgep_network.refresh_network_simple() RETURNS void SECURITY DEFINER AS $body$
```
m-kuhn, line 25: :+1:

m-kuhn (approved):

PR #172, opened 2021-02-04
followup fix for long running query when updating wastewater node symbology

olivierdalang:
**WARNING this alters a delta, merge before releasing 1.5.5**

Followup on https://github.com/QGEP/datamodel/pull/166 to deals with this case : https://github.com/QGEP/datamodel/pull/166#issuecomment-760749538

We now suspend the symbology triggers inside the update_wastewater_node_symbology function.

The issue was that some triggers are run FOR EACH ROW on all wastewater_structures when updating wastewater_node, resulting in a very slow query.

Ideally we should add conditions to only run these FOR EACH ROW triggers when specific columns are changed like this
```sql
CREATE TRIGGER ...
    ...
    FOR EACH ROW
    WHEN (OLD.somefield IS DISTINCT FROM NEW.somefield)
    EXECUTE PROCEDURE ...
```
but that's a bit out of scope of #166

### DIFF
--- a/06_symbology_functions.sql
+++ b/06_symbology_functions.sql
@@ -51,6 +51,14 @@ CREATE OR REPLACE FUNCTION qgep_od.update_wastewater_node_symbology(_obj_id text
   RETURNS VOID AS
   $BODY$
 BEGIN
+
+-- Otherwise this will result in very slow query due to on_structure_part_change_networkelement
+-- being triggered for all rows. See https://github.com/QGEP/datamodel/pull/166#issuecomment-760245405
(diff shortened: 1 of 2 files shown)


olivierdalang:
@ponceta Can you test and ensure this works as expected ? Should fix the issue you had with the long running query

ponceta:
Thanks for the fix, works as expected!

ponceta (approved):
LGTM

olivierdalang:
Thanks ! Merging this now just so that we don't forget to do so before release

PR #174, opened 2021-02-24
fix 632# part 2/2 : enforce identifiers in trigger

olivierdalang:
Fixes https://github.com/QGEP/QGEP/issues/632

This enforce population of identifier based on obj_id if empty.

This requires an update to pirogue : https://github.com/opengisch/pirogue/pull/10
Not sure this is really worth it hence separate PR

Do we want to apply the same logic for others views too ? Currently it's just done for structure parts.


olivierdalang:
Todo before merge :
- [ ] revert requirements.txt to use pypi package instead of github archive for pirogue

olivierdalang:
Currently failing due to https://github.com/opengisch/pirogue/issues/12

3nids (approved):

view/create_views.py
```diff
@@ -42,12 +42,12 @@ def create_views(srid: int,
 
     run_sql('view/drop_views.sql', pg_service, variables)
 
-    SingleInheritance('qgep_od.structure_part', 'qgep_od.access_aid', view_name='vw_access_aid', pg_service=pg_service, pkey_default_value=True).create()
```
3nids, line 45: the file permission change is causing troubles

PR #175, opened 2021-04-15
Qgep swmm2 update views

tproduit:


tproduit:
Update SWMM views to:
* solve https://github.com/QGEP/QGEP/issues/625
* solve https://github.com/QGEP/QGEP/issues/647
* Create two views used to show SWMM results in QGIS

sjib:
@tproduit does this also needs to be pushed?

PR #177, opened 2021-06-25
Create delta_1.5.6_kek_model_update_2008_2019.sql

sjib:
Adapting the differences between the qgep kek extension and the latest release of vsa-kek 2019. The following attributes were moved to the subclasses damage_channel and damage_manhole instead of damage: quantification1, quantification2, damage_begin, damage_end, distance (because MANDATORY in damage_channel but not in damage_manhole).

Also adding further changes to attributes and values of the KEK extension.


sjib:
@olivierdalang Can you help me with the checks failing - is this something I can influence or have to change or is this on the pipe something you would have to do. Thanks for a hint.

delta/delta_1.5.6_kek_model_update_2008_2019.sql
```diff
@@ -0,0 +1,424 @@
...
+
+
+-- delete columns in qgep_od.damage
+
+ALTER TABLE qgep_od.damage
+DROP COLUMN IF EXISTS quantification1,
+DROP COLUMN IF EXISTS quantification2,
+DROP COLUMN IF EXISTS damage_begin,
+DROP COLUMN IF EXISTS damage_end,
+DROP COLUMN IF EXISTS distance;
```
olivierdalang, line 74: These changes will require adapting the export script (not too difficult)

Are there other parts of QGEP that are affected by that (e.g. forms in the QGIS project) ?

delta/delta_1.5.6_kek_model_update_2008_2019.sql
```diff
@@ -0,0 +1,424 @@
...
+-- 1.5.6.3_kek_update_attribute_maintenance_event_result
+
+ALTER TABLE qgep_od.maintenance_event ALTER COLUMN  result TYPE varchar(255) ;
+COMMENT ON COLUMN qgep_od.maintenance_event.result IS 'Result or important comments for this event / Resultat oder wichtige Bemerkungen aus Sicht des Bearbeiters / Résultat ou commentaire importante de l''événement';
+
+-- 1.5.6.4_kek_add_vl_file_upddate_attribute_object
+
+ALTER TABLE qgep_od.file ALTER COLUMN  object TYPE varchar(16) ;
+COMMENT ON COLUMN qgep_od.file.object IS 'yyy_Objekt-ID (OID) des Datensatzes zu dem die Datei gehört';
+
```
olivierdalang, line 92: Not sure what these change are for ? (maybe you could hint the reason behind the change in the comment)

delta/delta_1.5.6_kek_model_update_2008_2019.sql
```diff
@@ -0,0 +1,424 @@
...
+-- 1.5.6.5_kek_add_vl_file_kind_add_new_values
+
+ INSERT INTO qgep_vl.file_kind (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (9146,9146,'digital_video','digitales_Video','video_numerique', 'video_digitale', 'rrr_digitales_Video', '', '', '', '', '', 'true');
+ INSERT INTO dictionary_od_values (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (9146,9146,'digital_video','digitales_Video','video_numerique', 'video_digitale', 'rrr_digitales_Video', '', '', '', '', '', 'true');
```
olivierdalang, line 96: As far as I remember, there's already an entry for digital_video, but spelled differently from the norm. If so, we should update instead of inserting

delta/delta_1.5.6_kek_model_update_2008_2019.sql
```diff
@@ -0,0 +1,424 @@
...
+
+-- set old value to false
+UPDATE qgep_vl.file_kind
+SET 	active = false
+WHERE  code = 3771; 
+
+UPDATE dictionary_od_values
+SET 	active = false
+WHERE  code = 3771; 
+
```
olivierdalang, line 111: Again not clear to me what the change is for, could you comment with the rationale ?

delta/delta_1.5.6_kek_model_update_2008_2019.sql
```diff
@@ -0,0 +1,424 @@
...
+ INSERT INTO qgep_vl.damage_channel_channel_damage_code (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (8907,8907,'BDEEB','BDEEB','BDEEB', 'BDEEB', 'BDEEB', '', '', '', '', '', 'true');
+ INSERT INTO dictionary_od_values (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (8907,8907,'BDEEB','BDEEB','BDEEB', 'BDEEB', 'BDEEB', '', '', '', '', '', 'true');
+ INSERT INTO qgep_vl.damage_channel_channel_damage_code (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (8908,8908,'BDEEC','BDEEC','BDEEC', 'BDEEC', 'BDEEC', '', '', '', '', '', 'true');
+ INSERT INTO dictionary_od_values (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (8908,8908,'BDEEC','BDEEC','BDEEC', 'BDEEC', 'BDEEC', '', '', '', '', '', 'true');
+
```
olivierdalang, line 423: Any idea why those were missing ?

olivierdalang (changes requested):
For the tests, there is a `CURRENT_VERSION.txt` file that must be updated to the latest version. 

This is currently the reason why tests are failing. Once this is fixed, it will (normally) still fail, as there will be differences between the deltas and the datamodel definitions, meaning we must also make the changes in the datamodel definition files (such as 04_vsa_kek_extension.sql).

In general, your changes mix model changes (eg. moving attributes from one table to another) with bugfixes (fixing errors/missing value lists).

Ideally we'd just do the bugfixes now for 1.5.6 so we don't introduce potential new issues and limit the work (no need to adapt the qgepqwat2ili script nor require updates to the QGIS project), and keep the rest for an upcoming release.

Otherwise we can do a backwards incompatible 1.6.0 with all the changes. For qgepqwat2ili I don't think it's much work, but if it requires updating the qgis project accordingly I'm not sure it's worth bothering all users just for that, so we could leave that for a future update when there's more things to change.

olivierdalang:
Also, there is another very small fix that would be needed for the ili export script :

```
ALTER TABLE qgep_od.reach_point ALTER CONSTRAINT rel_reach_point_wastewater_networkelement DEFERRABLE INITIALLY IMMEDIATE;
ALTER TABLE qgep_od.structure_part ALTER CONSTRAINT rel_structure_part_wastewater_structure DEFERRABLE INITIALLY IMMEDIATE;
```

This is because some inserts have cyclic dependencies, so that without deferring the constraints, we cannot insert at once (but need to insert the update the rows). Could you add that to this PR, so we keep everything in one place ?

delta/delta_1.5.6_kek_model_update_2008_2019.sql
sjib: Because the underlying norm of the VSA just introduced them with the update in 2019. https://vsa.ch/Mediathek/erhaltung-von-kanalisationen-2/ 
The Swiss norma again bases on an EN-Norm SN EN 13508-2 that added those new codes. Such changes normally reflext the need from daily work that then is integrated in the standard again.

delta/delta_1.5.6_kek_model_update_2008_2019.sql
sjib: The definitions of these attributes have changed e.g TEXT*41 instead of TEXT*20

delta/delta_1.5.6_kek_model_update_2008_2019.sql
sjib: To maintain the model so that you can see the changes also later, if a value changes during a model upgrade (patch or new release) it gets a new value ID. So I suggest to follow this principle als with the vl_* tables. You therefore have to add the new value ID, then replace old values with new values and then set the old value ID to active=false - see also below.

delta/delta_1.5.6_kek_model_update_2008_2019.sql
sjib: see comment above

delta/delta_1.5.6_kek_model_update_2008_2019.sql
sjib: I checked the project and found the following things that might be affected:

- [x] [vw_damage.yaml](https://github.com/QGEP/datamodel/blob/master/view/vw_damage.yaml) - it takes all attributes from super and subclass without specifying attribute names - so no problem
- [x] [vw_vw_maintenance_examination.yaml.yaml](https://github.com/QGEP/datamodel/blob/master/view/vw_maintenance_examination.yaml.yaml) - it takes all attributes from super and subclass without specifying attribute names - so no problem
- [ ] whether this influences the form for maintenance connected to it I do not know yet
- [ ] As you mentioned I have to change [04_vsa_kek_extension.sql](https://github.com/QGEP/datamodel/blob/master/04_vsa_kek_extension.sql) and
- [ ] [09_qgep_dictionaries_kek.sql](https://github.com/QGEP/datamodel/blob/master/09_qgep_dictionaries_kek.sql)
- [ ] I think then may be some of these fixes will not be needed anymore: https://github.com/QGEP/datamodel/tree/master/fixes
- [ ] if you see other points please add

sjib: I hope to have a new version for 04_vsa_kek_extension.sql and 09_qgep_dictionaries_kek.sql till tomorrow ready.

sjib: @olivierdalang I have pushed a new version of 04_vsa_kek_extension.sql and 09_qgep_dictionaries_kek.sql

olivierdalang:
(I've just added a commit to update current_version.txt, so that tests don't fail early)

olivierdalang:
I've made some changes to the github tests, they now fail successfully (haha).

To see actual reason of failure, click on the "Details" button next to a failed check (unless postgres/postgis version specific issues, it doesn't matter which version).  Then, unfold the "failure logs" step, and scroll to the very bottom.

Here you'll see the error that happened. In this case :

```
2021-07-13 13:55:56.851 UTC [246] ERROR:  column "fk_examination" of relation "damage" already exists
2021-07-13 13:55:56.851 UTC [246] STATEMENT:  ALTER TABLE qgep_od.damage ADD COLUMN fk_examination varchar (16);
```

Indeed, from the code it seems this is called twice, once added in your files, and once separately in `fix_vsa_kek_extension.sql` (not sure why it's separated ?!).


Also, I see with your last commit you added many many more changes in the definition that are not in the delta, which means we'll get a mismatch between the datamodel and the deltas, and tests will (normally) fail.

In order to have a working patch soon, I'd recommend :
- [ ] separating the changes : on one side urgent needs so that the ili export works (deferrable foreign keys and value_lists misspelling), and on the other side less urgent general vsa_kek improvement
- [ ] paying attention to the actual changes in the files, as to facilitate review (currently it seems many changes are actually just due to the order in the file, making review very difficult)

sjib:
@olivierdalang I have made the requested changes from the datamodel side - can you check on the remaining failures - I think they are not datamodel related anymore.

olivierdalang (approved):

sjib:
@olivierdalang Looks good to me.

sjib:
@olivierdalang Can you merge?

m-kuhn:
I think Olivier is on vacation, thanks for these updates!

PR #178, opened 2021-10-26
do not create set owner commands in plain dump

3nids:


sjib:
@3nids There are now two slq scripts for roles 12_0_roles.sql und 12_1_roles.sql: https://qgep.github.io/docs/installation-guide/database-initialization.html#create-minimal-roles-and-access

> The QGEP roles are defined in the https://github.com/QGEP/datamodel/blob/master/12_roles.sql file. It is recommended to use these when using QGEP in a production environment.

So do I have to install both?
https://github.com/QGEP/datamodel/blob/master/12_0_roles.sql
https://github.com/QGEP/datamodel/blob/master/12_1_roles.sql

Can you adjust the text?

Thanks

3nids:
@sjib done here https://github.com/QGEP/docs/commit/93dbb0dad11eef99f86c4859bd5c1b7f72f4aebf

PR #179, opened 2021-11-13
Fix labelling egg profile

sjib:
See https://github.com/QGEP/QGEP/issues/704


sjib:
Does it need an extra code to deal with when pp.height_width_ratio isNull ?

olivierdalang:
Looks fine to me, the ` pp.height_width_ratio is NULL` case seems already dealt with in the ELSE clause (it will fallback to `clear_height`)

olivierdalang (approved):

PR #182, opened 2022-04-11
Fix #733: Error when applying delta 1_5_6

sjib:
add drop vw_qgep_maintenance so there is not conflict in adding changes to qgep_od.maintenance_event

(related issue: QGEP/QGEP/issues/733)

### DIFF
--- a/delta/delta_1.5.6_kek_model_update_2008_2019.sql
+++ b/delta/delta_1.5.6_kek_model_update_2008_2019.sql
@@ -1,6 +1,8 @@
 -- Adapting first differences between the qgep kek extension and the latest release of vsa-kek 2019 for making INTERLIS import VSA-KEK plugin work.
 -- Full changes will be done in a next release (expected to be 1.6.0)
 
+-- 11.4.2022 add drop vw_qgep_maintenance so there is not conflict in adding changes to qgep_od.maintenance_event
+DROP VIEW qgep_od.vw_qgep_maintenance;
 
 -- Adding further changes to attributes and values
 -- 1.5.6.2_kek_update_attribute_maintenance_event_identifier - Changing from varchar (20) to varchar (41) so that longer identifiers can be imported as in testdata
--- a/delta/delta_1.5.6_kek_model_update_2008_2019.sql
+++ b/delta/delta_1.5.6_kek_model_update_2008_2019.sql
@@ -1,8 +1,8 @@
 -- Adapting first differences between the qgep kek extension and the latest release of vsa-kek 2019 for making INTERLIS import VSA-KEK plugin work.
 -- Full changes will be done in a next release (expected to be 1.6.0)
 
--- 11.4.2022 add drop vw_qgep_maintenance so there is not conflict in adding changes to qgep_od.maintenance_event
-DROP VIEW qgep_od.vw_qgep_maintenance;
+-- 11.4.2022 / 23.6.2022 add drop vw_qgep_maintenance IF EXISTS so there is not conflict in adding changes to qgep_od.maintenance_event
+DROP VIEW IF EXISTS qgep_od.vw_qgep_maintenance;
 
 -- Adding further changes to attributes and values
 -- 1.5.6.2_kek_update_attribute_maintenance_event_identifier - Changing from varchar (20) to varchar (41) so that longer identifiers can be imported as in testdata


sjib:
@olivierdalang Can you check on the docker errors e.g. https://github.com/QGEP/datamodel/runs/5978914622?check_suite_focus=true#step:15:6987

olivierdalang:
The error is ` view "vw_qgep_maintenance" does not exist`, which is caused by the `delta/delta_1.5.6_kek_model_update_2008_2019.sql` contained in your PR.

From the code, it seems that this view is not supposed to exist (I see `DROP VIEW IF EXISTS qgep_od.vw_qgep_maintenance;` in `delta_1.3.0_remove_old_views.post.sql`, so it should have been dropped already).

If you still have this view, it may indicate that you have an inconsistent database.

sjib:
@olivierdalang Can you review this?

olivierdalang:
@sjib 
I think I did already ? (see my comment above).

From what I understand, you're trying to delete something that does not exist in the datamodel, hence tests fail. IMO this PR should be closed without merge, as it rather looks like a problem of inconsistent database (so to be fixed on your DB) than a problem with the data model.

But maybe I miss some context (what exactly is it supposed to fix, how to you reproduce the issue, etc)

sjib:
DROP VIEW IF EXISTS qgep_od.vw_qgep_maintenance; instead of DROP VIEW as it might has been deleted already with delta_1.3.0_remove_old_views.post.sql

sjib:
@olivierdalang Thanks for the comment - I think now it should work well.

olivierdalang:
@sjib I don't think it's a good approach to include "IF EXISTS" statements in migrations. Migrations are supposed to migrate the database from a well known version to a well known version.

As you saw, `qgep_od.vw_qgep_maintenance` should have been deleted with `delta_1.3.0_remove_old_views.post.sql`. If the view is present in your database, it means that either  `delta_1.3.0_remove_old_views.post.sql` was never applied, or that this view was added afterwards manually. In both case, it means that your database is NOT in a consistent 1.5.5 version.

I think you must drop that view manually in your DB (and I recommend looking carefully for other inconsistencies with version 1.5.5, as they could lead to bugs or other migrations issues in the future).

Or am I missing some info as to why this view exist in your DB ?

sjib:
We had this original problem: https://github.com/QGEP/QGEP/issues/733

Applying delta_1.5.6_kek_model_update_2008_2019.sql Fehler "Typ einer Spalte, die von einer Sicht oder Regel verwendet wird, kann nicht geändert werden"
So it we found out that `qgep_od.vw_qgep_maintenance` is the problem. So for the average user it is very difficult to find out that `delta_1.3.0_remove_old_views.post.sql` has not been running. Or the database has been created with a later version.

And we still have https://github.com/QGEP/datamodel/blob/master/view/vw_maintenance_examination.yaml which creates that view - so can you check if it is still used in creation of a new database? 

And / or needed in any form (@urskaufmann you are more familiar with that).

olivierdalang:
Ahh it's more clear with the issue report (took the liberty to rename the PR and include the link to your description)

Is the intention:
A. to remove the vw_qgep_maintenance completely 
or
B. to fix delta 1.5.6
or
both ?

For A, you need to also prevent the view from being recreated by deleting https://github.com/QGEP/datamodel/blob/master/view/create_views.py#L61 (as well as the whole `vw_maintenance_examination.yaml` file)

For B, we cannot just edit the 1.5.6 migration (editing a past migration will make it be considered as non applied, until this is fixed: https://github.com/opengisch/pum/pull/81), so it would likely lead to other migration errors. We could drop the view in a pre-all step (this was already discussed a long time ago in https://github.com/QGEP/QGEP/issues/591, but not yet existing).

In short, no easy/quick fix, but instead we need to tackle some long overdue ground work...

sjib:
Trying to catch up on this:

1. @urskaufmann : Do we still need vw_qgep_maintenance in the current project? 
2. If no, I will adapt the create script as OD suggests instead of changing the 1.5.6 delta:

> For A, you need to also prevent the view from being recreated by deleting https://github.com/QGEP/datamodel/blob/master/view/create_views.py#L61 (as well as the whole vw_maintenance_examination.yaml file)

if yes then I will fix the 1.5.6 delta as PUM seems to be fixed and 81 is merged now and the arguments below are not valid anymore

> For B, we cannot just edit the 1.5.6 migration (editing a past migration will make it be considered as non applied, until this is fixed: https://github.com/opengisch/pum/pull/81), so it would likely lead to other migration errors. We could drop the view in a pre-all step (this was already discussed a long time ago in https://github.com/QGEP/QGEP/issues/591, but not yet existing). 

@olivierdalang @ponceta I see a general problem also in the future if we make changes in the datamodel to tables that are used in views and [then get this kind of error "Fehler "Typ einer Spalte, die von einer Sicht (View) oder Regel (Rule) verwendet wird, kann nicht geändert werden""](https://github.com/QGEP/QGEP/issues/733) when we do not have a process during PUM that 

- first deletes all views, 
- then applies all deltas and 
- in the end recreates all views.

Or does PUM work already work like that? Thanks for clarification.

urskaufmann:
Do we still need vw_qgep_maintenance in the current project?
I think its not worth to change anything with this view in version 2015. We have to change with DSS 2020 anyway. So it's "Yes".

sjib:
Not needed anymore if https://github.com/QGEP/datamodel/pull/146 is integrated, so that PUM works like expected:
-     first deletes all views,
-     then applies all deltas and
-     in the end recreates all views.

Set priority to implement https://github.com/QGEP/datamodel/pull/146

PR #184, opened 2022-08-11
bump PUM to 0.10.0 and adapt accordingly

olivierdalang:
This runs CI with the new changes in PUM (deltas by filename instead of checksum, improved pum check, etc.).

The new version of PUM highlighted some inconsistencies which should be fixed by `delta/delta_1.5.7_fix_inconsistencies.sql`.

Before merging:
- [x] make a release of PUM and adapt requirements.txt accordingly (not yet on pypi, rerun tests once published)


3nids (approved):

olivierdalang:
I do not think so.

The error you show looks related to postgres versions (trying to restore a postgres>=10 dump to a postgres<10 database).

PR #186, opened 2022-09-22
Labeling of wastewater structures with multiple wastewater nodes

cymed:
According to VSA-DSS, a wastewater structure with a leaping weir is represented with two wastewater nodes see [VSI-Wiki](https://vsa.ch/wiki/regenueberlauf-mit-streichwehr/). This requires us to partition the input and output labels by NE.fk_wastewater_structure instead of RP.fk_wastewater_networkelement. 

Additionally, I added two other enhancements:

1. Covers are enumerated
2. The enumeration index is only added if there is more than one input/output/cover.

PR #187, opened 2022-10-11
Adjust vw qgep overflow

cymed:
As the wastewater node geometries are 3D, the Linestrings should be too. otherwise I get the error 
```
ERROR:  Geometry has Z dimension but column does not 
SQL state: 22023
```


sjib:
@cymed Can you add a delta also for this change in the symbology functions - see https://github.com/QGEP/datamodel/tree/master/delta for examples

sjib:
I can run this manually only, if I delete the on_reach_update trigger first.
Would we have to include a delete and re-create of all triggers also or are they anyway included with the views?

See https://github.com/QGEP/QGEP/issues/798 and Pullrequest https://github.com/QGEP/datamodel/pull/146

I think we would have a least to run drop_symbology triggers first also with delete views?
https://github.com/QGEP/datamodel/blob/6aa60794729f24edcd9a7c6d7ad124a605dec4e9/06_symbology_functions.sql#L648

ponceta:
Here we need add the modification to the symbology function to the delta and remove the view and driggers from it since it should be already taken care of through the view recreating.

My only interrogation is the geometry type of this vw_qgep_overflow? If it is a LinestringZ the definition has to be adapted.

cymed:
> Here we need add the modification to the symbology function to the delta and remove the view and triggers from it since it should be already taken care of through the view recreating.
> 
> My only interrogation is the geometry type of this vw_qgep_overflow? If it is a LinestringZ the definition has to be adapted.
> 
> https://github.com/QGEP/datamodel/pull/187/files#diff-6d148c192cfb1fae5eabe0e90fae5614c09faca6803ff6982efd6d401efa9937L6-R6

The idea of the view is to use it inside QGIS (in a later update of the .qgs) to create overflows. Overflows are pseudo-linear, but for visualisations the use of a point is preferable:

- Easier to visualise on large scales
- We can't update the geometry of the Linestring in QGIS, as the geometry is calculated by the view

urskaufmann:
I like this view. If the project is configured, it's easy to work with it.
1. Symbol Regenüberlauf is not from overflow, but from the wastewater-structure (function Regenüberlauf).
2. With vw_qgep_overflow, I find the overflows and it visualises from where to where the overflow is. Normally you can see the line only in a detail scale.
3. This ist much easier than to search in a tab of the wastewater-node, if there exists an overflow
4. The create a new overflow, it is very easy to create the new record, then choose the node and the to-node on the map and that's it (without first opening the form-window of the wastewater-structure)

cymed:
@ponceta the delta from #186 is more recent than this one. We only need to change the vw_oo_overflow.yaml

ponceta:
This requires a subsequent .qgs project adaptation.

view/vw_oo_overflow.yaml
```diff
@@ -3,10 +3,9 @@ table: qgep_od.overflow
 view_name: vw_qgep_overflow
 
 additional_columns:
-  geometry: ST_MakeLine(n1.situation_geometry, n2.situation_geometry)::geometry('LineString',%(SRID)s)
+  geometry: n1.situation_geometry::geometry('PointZ',%(SRID)s)
```
sjib, line 6: This is now a point instead of a line - I am not sure if that is what @urskaufmann intended this to be.

sjib: @cymed can you check with @urskaufmann if this still is what both of you expect of this?

cymed: @sjib: @urskaufmann and I agreed to changing the geometry from a line to a point in November. I can forward you the discussions by mail.

sjib: OK, then everything is ok

PR #189, opened 2022-11-09
Create `vw_infiltration_installation` and `vw_wwtp_structure` views

rbovard:
Fix https://github.com/QGEP/QGEP/issues/768

Create `vw_infiltration_installation` and `vw_wwtp_structure` views.

~Please double check the diff file, I created it manually.~

~*BTW, is there a way to automatically create this diff (maybe with PUM)?*~


ponceta:
@rbovard Nope nope nope, no delta is required on views, that's precisely why we drop and create these dynamically.

Without the delta CI should be fine!

sjib:
@rbovard My fault with the delta. Just for datamodel changes.

ponceta:
CI is happy, safe to merge since this is only two additionnal views.

ponceta:
Thank you @rbovard !

PR #191, opened 2022-12-06
add views for VSA-DSS subsystem property drainage

cymed:
In order to make full use of the DSS-subsystem property drainage (Liegenschaftsentwässerung) we need the corresponding views


ponceta:
Small enhancement to be added to the .qgs project https://github.com/QGEP/QGEP/issues/772

rbovard:
Shouldn't they also be added in [drop_views.sql](https://github.com/QGEP/datamodel/blob/master/view/drop_views.sql)?

sjib:
> Shouldn't they also be added in [drop_views.sql](https://github.com/QGEP/datamodel/blob/master/view/drop_views.sql)?

Yes they should, as you did with [your new views](https://github.com/QGEP/datamodel/pull/189/files). Thanks for the hint @rbovard 

@cymed Can you update  [drop_views.sql](https://github.com/QGEP/datamodel/blob/master/view/drop_views.sql) with an addtional pull request?

PR #192, opened 2022-12-08
Interpolate z of reaches' progression_geometry

cymed:
fixes [#759](https://github.com/QGEP/QGEP/issues/759)

### DIFF
--- a/14_geometry_functions.sql
+++ b/14_geometry_functions.sql
@@ -87,3 +87,87 @@ CREATE TRIGGER synchronize_level_with_altitude
 -- Synchronize reach and reach_point GEOMETRY with level qgep_od.vw_qgep_reach is implemented in vw_qgep_reach.sql
 -----------------------------------------------
 -----------------------------------------------
+
+
+
+-----------------------------------------------
+-----------------------------------------------
+-- Interpolate z value for reach GEOMETRY
+-----------------------------------------------
+-----------------------------------------------
+CREATE OR REPLACE FUNCTION qgep_sys.re_interpolate_vertex_z(
+	re_obj_id character varying)
+    RETURNS geometry
+    LANGUAGE 'plpgsql'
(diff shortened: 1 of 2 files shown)


cymed:
Downside of this solution: does not work with reaches that have duplicate nodes

sjib:
Duplicate nodes in a reach should be eliminated anyway - will cause problems on the INTERLIS export.

ponceta:
Should be tested for durability issues like :
https://github.com/QGEP/datamodel/blob/master/test/test_triggers.py

cymed:
Docker is failing due to the :SRID in the delta. How can I fix this?

olivierdalang:
> Docker is failing due to the :SRID in the delta. How can I fix this?

In the delta, you should use `%(SRID)s` instead (that variable will be set by PUM).

olivierdalang (changes requested):
Some notes:

- [ ] in line with what we're doing in https://github.com/QGEP/datamodel/pull/146, we should rather drop/recreate functions in pre/post all, instead of creating a delta. Otherwise, any slight modification of the function will result in a delta. This means the code for the function would move under `./views` (which already not only contains views)

- [ ] to me, it's not clear what the goal of this PR is. It creates a new function, but I don't see it used anywhere ? Or am I missing something ? If it's meant as a helper function to be run manually, we probably at least would require some docs. If not you're probably missing triggers.

- [ ] if meant as a helper function, it would be better to implement this as a QGIS algorithm, otherwise the datamodel will quickly be cluttered with util functions that need to be maintained. I'd recommend searching first if by any chance it does not already exist in an existing plugin

cymed:
It is not used anywhere at the moment, correct. There are discussions going on whether to include it in a trigger or not. Until then, I suggest to close this PR and reuse the code once the trigger functionalities are defined.

PR #193, opened 2022-12-08
2022 12 fix vsa kek file fk datamedia relation

sjib:
Add missing relation for file -> data_media


sjib:
@ponceta Can you check and define a delta name for this?

04_vsa_kek_extension.sql
```diff
@@ -959,6 +958,8 @@ ALTER TABLE qgep_vl.data_media_kind ADD CONSTRAINT pkey_qgep_vl_data_media_kind_
  ALTER TABLE qgep_od.data_media ADD CONSTRAINT fkey_vl_data_media_kind FOREIGN KEY (kind)
  REFERENCES qgep_vl.data_media_kind (code) MATCH SIMPLE 
  ON UPDATE RESTRICT ON DELETE RESTRICT;
+ALTER TABLE qgep_od.file ADD COLUMN fk_data_media varchar(16);
+ALTER TABLE qgep_od.file ADD CONSTRAINT rel_file_data_media FOREIGN KEY (fk_data_media) REFERENCES qgep_od.data_media(obj_id) ON UPDATE CASCADE ON DELETE set null;
```
rbovard, line 962:
```suggestion
ALTER TABLE qgep_od.file ADD CONSTRAINT rel_od_file_data_media FOREIGN KEY (fk_data_media) REFERENCES qgep_od.data_media(obj_id) ON UPDATE CASCADE ON DELETE set null;
```

rbovard (changes requested):
I suggested a new name for the constraint to be consitent with the other ones.

For the `DELETE` rule, I don't know was is the best according to the norms, I can't pronounce myself about that.

04_vsa_kek_extension.sql
sjib: The general idea is good, but currently only the fk_dataowner / fk_provider follow this logic. 

The other relations between classes still follow this logic:
ALTER TABLE qgep_od.water_catchment ADD CONSTRAINT rel_water_catchment_aquifier FOREIGN KEY (fk_aquifier) REFERENCES qgep_od.aquifier(obj_id) ON UPDATE CASCADE ON DELETE set null;
https://github.com/QGEP/datamodel/blob/e94fef6296354464e5855be98275596af0b0c3aa/03_qgep_db_dss.sql#L3539

delta/delta_1.5.x_kek_model_missing_relation_file.data_media.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE qgep_od.file ADD CONSTRAINT rel_file_data_media FOREIGN KEY (fk_data_media) REFERENCES qgep_od.data_media(obj_id) ON UPDATE CASCADE ON DELETE set null;
```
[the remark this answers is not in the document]
sjib: same see above

04_vsa_kek_extension.sql
rbovard: Oh I see, sorry I didn't noticed that (I was focused on the table ;)).

sjib:
The comment of @rbovard lists some inconsistencies - but the 
`ALTER TABLE qgep_od.examination ADD CONSTRAINT rel_examination_reach_point FOREIGN KEY (fk_reach_point) REFERENCES qgep_od.reach_point(obj_id) ON UPDATE CASCADE ON DELETE set null;`
is also without od, dito many other relations in https://github.com/QGEP/datamodel/blob/master/03_qgep_db_dss.sql
except the datenherr / provider relations.

We have here an iconsistency, but for the new relation I think it is correct

sjib:
@ponceta Can you provide a delta id for this delta_1.5.x_kek_model_missing_relation_file.data_media.sql, so I can add also update `system/CURRENT_VERSION.txt`
I have added this bugfix to the roadmap: https://github.com/orgs/teksi/projects/2/views/1?pane=issue&itemId=18886384

ponceta:
Please rename delta to 1.5.7

PR #194, opened 2022-12-08
Handle null data_media path for file url

rbovard:
Fix part of https://github.com/QGEP/QGEP/discussions/780 *(point 3)*


ponceta:
Much better, thank you! @rbovard

PR #196, opened 2022-12-21
Update of drop_views for property drainage

cymed:
See #191


rbovard:
I think there is a typo: `vw_vw_xxx`

And why not put them in the same place as in you did [here](https://github.com/QGEP/datamodel/pull/191/files)?

ponceta:
@rbovard I don't understand what you mean by the same place?

PR #197, opened 2023-01-06
Add _status to Wastewater node

cymed:
As catchments should only be linked to wastewater nodes whose _function_hierarchic is pwwf.* and whose wastewater structure's status is nor abandoned nor inoperative, it would be practical to add the wastewater structure's status to the wastewater node in QGEP.


ponceta:
A nice enhancement👍 

Could you provide the delta to that?

PR #199, opened 2023-02-20
Fix symbology for nodes

ponceta:
Fix symbology for nodes by adding the _all option missing in the WHERE clause


cymed (commented):
Looks good to me, thanks

PR #200, opened 2023-02-23
[WIP] Update docker tests to future postgres postgis versions

ponceta:
Update docker tests to actual postgres postgis versions

Use ubuntu LTS check old postgres postgis, soon to be old, recent and newest


olivierdalang:
Which release does this PR target ? We have no defined policy for 1.x.x, but in the future, dropping support for postgres versions changes would have to be frozen for the next yearly release, so IMO we should already apply this policy and freeze that for 2023.0.

ponceta:
@olivierdalang I was only checking automated tests and CI options but the goal is to change that for the 2023.0 Release once the 1.6.0 step is finished. Release policy is in what we discussed a part of 2023 release.

https://github.com/orgs/teksi/projects/2/views/4?pane=issue&itemId=21154279

ponceta:
Fix warnings on deprecated libraries and actions :

image

ponceta:
@3nids it looks like 3.2 and 3.4 have been fixed but not 3.3, any hint / good practices on how to prevent users from using it?

3nids:
good question, the simplest is probably in the SQL install script to detect `SELECT PostGIS_full_version()` and raise an error.

sjib:
@ponceta Is this still wanted? If yes can you make it 1.6.2 ready

ponceta:
Postgis 3.3 docker image has not been updated so it is not fixed, postgis 3.2 is available on postgres 14-3.2 but is already two years old.

ponceta:
`system_user ` now looks like a reserved keyword, we have to adapt it to support postgres 16.

ponceta:
@sjib here is the full error log if required for postgres 16

sjib:
https://github.com/QGEP/datamodel/pull/238 should help to solve this.

sjib:
@ponceta Still an issue? Do you want this also fixed with the next release? Then please add to milestone 1.6.4

PR #203, opened 2023-02-23
Reapply #197 and followup

olivierdalang:
This re-applies commits reverted in https://github.com/QGEP/datamodel/pull/201 (about #197 and followups).
This depends on #201 which should be merged first.

This PR needs to be fixed and have CI pass before merge.

@cymed and @ponceta as the original co-authors of #197, are you able to take it from here and make the necessary changes until CI passes ?

(note: it seems CI is not triggered yet, probably because this does not yet target the master branch. CI should run (and fail) as soon as #201 is merged)

### DIFF
--- a/05_data_model_extensions.sql
+++ b/05_data_model_extensions.sql
@@ -78,3 +78,7 @@ ALTER TABLE qgep_od.wastewater_node ADD COLUMN _function_hierarchic integer;
 COMMENT ON COLUMN qgep_od.wastewater_node._function_hierarchic IS 'not part of the VSA-DSS data model
 added solely for QGEP
 has to be updated by triggers';
+ALTER TABLE qgep_od.wastewater_node ADD COLUMN _status integer;
+COMMENT ON COLUMN qgep_od.wastewater_node._status IS 'not part of the VSA-DSS data model
+added solely for QGEP
+has to be updated by triggers';
--- a/06_symbology_functions.sql
+++ b/06_symbology_functions.sql
@@ -59,28 +59,34 @@ IF _all THEN
   PERFORM qgep_sys.drop_symbology_triggers();
 END IF;
 
+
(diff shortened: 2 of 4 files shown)


olivierdalang:
By the way, just in case it wasn't clear, the place to look out for error messages in the CI (when they happen during the datamodel initialisation)  is here :
image

In this specific case, you will see:
```
     [many many more output lines above]
     Applying delta 1.4.0 swmm_views... OK
     Applying delta 1.4.0 update_wastewater_structure_label_with_bottom_level... OK
     Applying delta 1.4.1 swmm_catchment_areas... OK
     Applying delta 1.5.0 fix_network_tracking_views... OK
     Applying delta 1.5.1 fix_network_tracking_views... OK
     Applying delta 1.5.2 fix_network_tracking_views... OK
     Applying delta 1.5.3 fix_symbology_function... OK
     Applying delta 1.5.3 fix_vl_value_en_digital_video... OK
     Applying delta 1.5.3 fix_ws_output_labels... OK
     Applying delta 1.5.5 add_spatial_index_on_network_views... OK
     Applying delta 1.5.5 fix_dictionary_od_table_datenlieferant... OK
     Applying delta 1.5.5 fix_permissions_on_network_refresh_function... OK
     Applying delta 1.5.5 symbology_for_nodes... OK
     Applying delta 1.5.6 kek_model_update_2008_2019... OK
     Applying delta 1.5.7 add_status_to_wn... syntax error at or near "FUNCTION"
LINE 38:     EXECUTE FUNCTION qgep_od.ft_vw_wastewater_node_delete();
```

ponceta:
Thank you @olivierdalang !

cymed:
@ponceta Are the deltas applied before the creation of views? if so, all is good (because create_views.py will create the views etc. correctly). If not I propose adding another commit to include _status in the view (as we want the _status value in QGIS)

PR #204, opened 2023-02-24
bump actual version to target version 1.6.0 for imminent release

ponceta:
bump actual version to target version 1.6.0 for imminent release


ponceta:
@olivierdalang does this fail because I do not propose any deltas?

olivierdalang:
> @olivierdalang does this fail because I do not propose any deltas?

Yes. I guess for now the easiest would be to create a noop delta (`SELECT 1;`), so we get the correct name in the database.

ponceta:
Datamodel is ready for 1.6.0 release

PR #205, opened 2023-02-28
Run drop views extended to network views too

ponceta:
Run drop views extended to network views too


ponceta:
@olivierdalang you can see here the changes in the delta

https://github.com/QGEP/datamodel/pull/205/files#diff-5ab0163b1ab00a16c0718297fb751405f18f9b41d1fd56dd876af7318d5e9872L2-L3

Since theses views are materialized views, they should be refreshed at least every day in a production environment.

If they are recreated with the upgrade process, this would ensure that they are at least refreshed once per year or more with a rolling release. 

My issue was that theses views were not recreated after the Coordinate System reference system since they were not appearing in the view folder.

This commits avoids having code about theses views in the deltas for the future too.

PR #206, opened 2023-03-03
Change attributes class to classname to avoid issues with the reserved keyword class in python

sjib:
Should solve [Reserved key word class in python](https://github.com/QGEP/qgepqwat2ili/issues/72)


ponceta:
Now you can look at the failure logs as explained here :

https://github.com/QGEP/datamodel/tree/master/CI

```
2023-03-06 07:15:10.152 UTC [281] ERROR:  column "class" of relation "qgep_od.txt_symbol" does not exist
2023-03-06 07:15:10.152 UTC [281] STATEMENT:  COMMENT ON COLUMN qgep_od.txt_symbol.class IS 'ID of the active zone, for visualization purpose (use A, B, C, D, E, F, G and H)';
```

sjib:
maintenance_event.active_zone is a TEKSI extension to VSA-DSS and should be renamed to maintenance_event.xxx_active_zone. I think this was introduced by Koni.

See https://qgep.github.io/docs/admin-guide/db-customization/index.html#adding-fields

Does anyone still use this? @urskaufmann @TriformMoritz @varrieta @ponceta

urskaufmann:
never used this field...

PR #208, opened 2023-03-08
[WIP] Allow vl extensions for swmm views

cymed:
For those who service data models other than VSA, value list extensions might be needed (i.e. AG-96 in the Canton of Aargau, which adds additional function_hierarchic for channels and drainage_system_* for catchment areas ). If one uses such extensions, the SWMM views won't work, as the hard-coded where clauses of the views have a different scope of options. By using the vsacode instead of the code in qgep_vl.*, these problems can be solved elegantly.

The pull request handles only the view qgep_swmm.vw_junctions so far. If the conceptual change is approved by @tproduit, I will extend oh the pull request.


ponceta:
To be discussed with @tproduit directly but I see this in a good way, meaning it is the datamanager responsability to fill the vsa_code for every of its custom values. This would be a good practice in general when dealing with custom values and could/should be documented in that way.

tproduit (approved):

tproduit:
It seems also ok for me.

cymed:
Tests for export from a Database with vl extensions are pending

cymed:
closing pull request from my master branch, new Branch is #220

PR #217, opened 2023-04-13
change identifier length for maintenance event

cymed:
Maintenance event identifiers in VSA DSS 2015 have a maximum length of 20 characters , in TEKSI wastewater, up to 41 characters are allowed


urskaufmann:
similar: Maintenance event result (Ergebnis) in VSA DSS 2015 has a maximum length of 50 characters , in TEKSI wastewater, up to 255 characters are allowed

urskaufmann:
This difference in length is since datamodel release 1.5.6. There is a difference between datamodel dss_v2015 and kek_2019 in the length of this two fields in maintenance_event.
The problem is: I can not export my 1.5.6 database with dss_2015. Seems this is the reason for my problem with import dss in the qgep-plugin test https://github.com/QGEP/qgepqwat2ili/issues/118

sjib:
List of difference VSA-KEK 2019/ 2020 and qgep see https://github.com/QGEP/qgepqwat2ili/issues/91#issuecomment-1514785218

cymed:
closing branch to prevent possible duplicates

PR #220, opened 2023-06-20
Allow vl extensions in SWMM views

cymed:
As [#208](https://github.com/QGEP/datamodel/pull/208) was accidentally pushed from my master branch, i had to do some cleanup. Same changes in swmm views, but not from master


cymed:
Tested the export on a database with vl extensions. All errors are not related to the vl extension.

ponceta:
@ponceta review and test to do

sjib:
Could also go in 1.6.1 if ok

PR #225, opened 2023-09-05
Include overflows in _usage_current

cymed:
There are several situations (i.E. a stormwater composite tank with a preceding separating structure) where a wastewater structure only has overflows as topological connections. In these situations, we should also set a _usage_current dependent on the incoming overflow (outgoing overflows pose problems for pumps). This PR includes them in the updating process for wastewater structures and wastewater nodes.


sjib:
@cymed You should adapt this also in 06_symbology_functions.sql - not just in a delta (or am I wrong @ponceta ?)
https://github.com/QGEP/datamodel/blob/8a64cdbe15982ba3f490c9e4e4223a3e44c8fce7/06_symbology_functions.sql#L50

ponceta:
@ponceta Needs a small test before getting into core 1.6.1

cymed:
Imho we can close this pr and focus on TWW @ponceta

sjib:
If it is working I would still include it in qgep

ponceta (approved):

PR #228, opened 2023-10-16
Enable Multitenancy

cymed:
As a first step to multitenancy, we should define how to handle baskets. This PR is a suggestion for the architecture. 

TO DO:

- [ ] Alter all views
- [ ] Update qgepqwat2ili

Will need a Feature request to be fully implemented


cymed:
Docker fails because the existing function cannot be dropped.

PR #229, opened 2023-10-17
add additional structure part views

cymed:
This PR adds the structure part views that are included in VSA DSS 2015 but not SIA 405


sjib (approved):
Looks good to me. With 2020 there will be one more: Spuelstutzen (flushing_nozzle)

PR #236, opened 2024-06-19
Create vw_reaches_with_blank_connections.sql

sjib:
View that shows all reaches with blank connections (Blindanschlüsse). Can be used for visualisation
Author Urs Kaufmann 19.6.2024 
@urskaufmann OK to publish?


urskaufmann (approved):

ponceta:
This will never be called, if that's OK for you feel free to merge, if you want to have it natively added to the datamodel it has to be included to the initialization scripts.

PR #237, opened 2024-06-20
Add OID for Noville

rbovard:


cymed:
I think it is smarter to add OID prefixes on the local database instead of the github repo

rbovard:
@cymed I did the same as for the other prefixes (thanks @ponceta for the merge).

But if you want to change this, it will be better to remove all existing prefixes to avoid confusion.

cymed:
I will do so in tww

PR #238, opened 2024-08-22
2024 08 22 rename mutation.system user to avoid postgres confllicts

sjib:
should solve part of https://github.com/QGEP/QGEP/issues/862


sjib:
@ponceta how to adapt here?
  
C:\Daten\QGEP\github\datamodel\test_data\qgep_demodata_1.0.0.dump (4 Treffer)
	Zeile  4921:     system_user character varying(20),
	Zeile  4937:

sjib:
@ponceta Thanks for the delta!

ponceta (approved):

rbovard:
Thank you for the fix! We encountered this issue while restoring a dump in PostgreSQL 16.

FYI this can also be resolved by using quotes around the reserved column name.

ponceta:
@rbovard Thank you for the feedback! I have no functional CI for postgres 16 so be careful and tell us if you encounter anything new!

PR #241, opened 2024-11-15
Create .pre-commit-config.yaml

ponceta:


ponceta:
@domi4484 any idea how to make the CI happy about this import? :

https://github.com/QGEP/datamodel/pull/241/files#diff-c883dd6eb4fbc752e5147566aea0c5a02a08897683b2f80005dde8f7fcae614eR11-R14

in 
`create_views.py`
`pre-all.py`
`post-all.py`

They are launched by pum in the CI so I don't get how to reach the view classes.

domi4484 (approved):
I guess the only way is to make flake8 ignore the rule in this case.

Not sure what you mean about "launched by PUM", are those file generated autmatically?

ponceta:
> I guess the only way is to make flake8 ignore the rule in this case.
> 
> Not sure what you mean about "launched by PUM", are those file generated automatically?

Theses files are provided with a specific syntax so PUM applies these before and after the deltas when upgrading :

https://github.com/opengisch/pum?tab=readme-ov-file#delta-files

domi4484 (commented):
Ok thanks for the insight.

I guess the noqa in fact must be on the import lines

PR #244, opened 2025-01-21
Alter postgres to the 2025 version

ponceta:
Alter postgres to the 2025 version

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG POSTGIS_VERSION=14-3.2
+ARG POSTGIS_VERSION=16-3.5
 FROM postgis/postgis:${POSTGIS_VERSION}
 
 # System deps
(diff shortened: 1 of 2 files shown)


ponceta:
system_user prevents us from upgrading to postgreSQL 16 + since it is only fixed by deltas.

@3nids @domi4484 What would be the simplest thing to do? Upgrade process to a v2.0 then start from v2.0 until end of life of qgep?

3nids:
To me the simplest and safest would be to alter the initial dump, to either double quote system_user (it might work, not sure and didn't check) or directly rename the column there. Then, adapt the delta to correctly handle both scenarios (system_user and user_system).

Another approach would be to simply drop these checks. And only go with deltas from now on. Who's going to init a new instance on QGEP anyway? he would go for TWW directly. I would recommend following this road.

sjib:
@ponceta Still needed or solved with TWW Migration?

PR #246, opened 2025-09-25
Fix requirements.txt

ponceta:
Otherwise pum 1.x is installed by default which is not supported

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pirogue==1.4.1
 psycopg2
 pyyaml
-pum>=0.10.0
+pum<=0.10.0


sjib:
good hint!
@ponceta did you check whether their is also a conflict with the roles definition between qgep and tww? Would it be good practice to install tww in another database cluster than qgep?

ponceta:
Until yet I've seen no trouble linked to postgreSQL side. Python packaging is always a bit tricky and especially on Windows where it is poorly supported and often conflicting. It would be easier to ship pirogue with tmmt directly as we do with pum but it has a limitation which implies that we upgrade pirogue for all modules and versions at the same time.

